PR #37, opened 2018-03-20
refactor(*): refactor agents

jeyrschabu:
- added an agentRunner abstracting locking mechanism
- refactor how run age is calculated


jeyrschabu:
FYI @asher 
This simplifies how agents are run.

PR #38, opened 2018-03-20
chore(metrics): Adding metrics

jeyrschabu:
- recording a number of the handlers metrics

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
@@ -16,8 +16,10 @@
 
 package com.netflix.spinnaker.swabbie
 
+import com.netflix.spectator.api.Id
 import com.netflix.spectator.api.LongTaskTimer
 import com.netflix.spectator.api.Registry
+import com.netflix.spinnaker.swabbie.events.Action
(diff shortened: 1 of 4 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceHandler.kt
```diff
@@ -144,12 +147,50 @@ abstract class AbstractResourceHandler<out T : Resource>(
...
+                                workConfiguration: WorkConfiguration,
+                                violationCounter: AtomicInteger,
+                                candidateCounter: AtomicInteger) {
+    markDurationTimer.stop(markerTimerId)
+    registry.gauge(
+      candidatesCountId.withTags(
+        "resourceType", workConfiguration.resourceType,
+        "configuration", workConfiguration.namespace,
+        "resourceHandler", javaClass.simpleName,
+        "violations", violationCounter.get().toString()
```
asher: I steered you wrong suggesting setting storing violation count as a tag - this is actually a bad idea as it results in a new timeseries for each unique value of `violationCounter`.

The easiest fix would be to store a separate metric for violations. A way to also store violation or other counts within `swabbie.resources.candidatesCount` without blowing up the backend would be to have something like a "status" tag with values like "marked" and "violations". Then update the gauge in separately for each status value, so that when `"status", "violations"`, you'd `set(violationCount.get())`. I think having a separate `swabbie.resources.markViolations` metric would make more sense though to someone browsing metrics without knowledge of the code.

PR #44, opened 2018-03-30
refactor(*): Typifying owner resolver

jeyrschabu:
- added type def to resource owner resolver


jeyrschabu:
Per our convo @christopherthielen, I updated the owner resolution bits passing the type down.

PR #58, opened 2018-07-10
feat(Images): Adding support for scanning Images

jeyrschabu:
- added `OrphanedImageRule` to look for unused images
- rule looks to see if image is referenced by instances & launch configs
- looks to see if the current image has siblings in other accounts

### DIFF
--- a/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
+++ b/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
@@ -1,0 +1,269 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.swabbie.aws.images
+
+import com.netflix.spectator.api.Registry
+import com.netflix.spinnaker.kork.core.RetrySupport
+import com.netflix.spinnaker.moniker.frigga.FriggaReflectiveNamer
+import com.netflix.spinnaker.swabbie.*
+import com.netflix.spinnaker.swabbie.aws.autoscalinggroups.checkStatusDelay
+import com.netflix.spinnaker.swabbie.aws.instances.AmazonInstance
+import com.netflix.spinnaker.swabbie.aws.launchconfigurations.AmazonLaunchConfiguration
+import com.netflix.spinnaker.swabbie.echo.Notifier
+import com.netflix.spinnaker.swabbie.exclusions.ResourceExclusionPolicy
+import com.netflix.spinnaker.swabbie.model.*
+import com.netflix.spinnaker.swabbie.orca.OrcaJob
+import com.netflix.spinnaker.swabbie.orca.OrcaService
+import com.netflix.spinnaker.swabbie.orca.OrchestrationRequest
+import kotlinx.coroutines.experimental.Deferred
+import kotlinx.coroutines.experimental.async
+import kotlinx.coroutines.experimental.delay
+import kotlinx.coroutines.experimental.runBlocking
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.stereotype.Component
+import java.time.Clock
+import java.util.*
+import kotlin.system.measureTimeMillis
+
+@Component
+class AmazonImageHandler(
+  registry: Registry,
+  clock: Clock,
+  notifier: Notifier,
+  resourceTrackingRepository: ResourceTrackingRepository,
+  resourceOwnerResolver: ResourceOwnerResolver<AmazonImage>,
+  exclusionPolicies: List<ResourceExclusionPolicy>,
+  applicationEventPublisher: ApplicationEventPublisher,
+  lockingService: Optional<LockingService>,
+  retrySupport: RetrySupport,
+  private val rules: List<Rule<AmazonImage>>,
+  private val imageProvider: ResourceProvider<AmazonImage>,
+  private val instanceProvider: ResourceProvider<AmazonInstance>,
+  private val launchConfigurationProvider: ResourceProvider<AmazonLaunchConfiguration>,
+  private val accountProvider: AccountProvider,
+  private val orcaService: OrcaService
+) : AbstractResourceTypeHandler<AmazonImage>(
+  registry,
+  clock,
+  rules,
+  resourceTrackingRepository,
+  exclusionPolicies,
+  resourceOwnerResolver,
+  notifier,
+  applicationEventPublisher,
+  lockingService,
+  retrySupport
+) {
+  override fun deleteMarkedResource(markedResource: MarkedResource, workConfiguration: WorkConfiguration) {
+    orcaService.orchestrate(
+      OrchestrationRequest(
+        application = FriggaReflectiveNamer().deriveMoniker(markedResource).app ?: "swabbie",
+        job = listOf(
+          OrcaJob(
+            type = "deleteImage",
+            context = mutableMapOf(
+              "credentials" to workConfiguration.account.name,
+              "imageId" to markedResource.resource.resourceId,
+              "cloudProvider" to markedResource.resource.cloudProvider,
+              "regions" to listOf(workConfiguration.location)
+            )
+          )
+        ),
+        description = "Deleting Image Id:${markedResource.resourceId} Name: ${markedResource.name}"
+      )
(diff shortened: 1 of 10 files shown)


jeyrschabu:
Configuration snippet to enable image handling:
```
swabbie:
  .
  .
  providers:
    - name: aws
      locations:
        - us-east-1
      accounts:
        - test
      resourceTypes:
        - name: image
          enabled: true
          dryRun: false
          retentionDays: 14
```
This can be run in `dryRun` mode by toggling the flag in the config.
If you wish to whitelist images during testing, you can do so via exclusions:
```
- name: image
  enabled: true
  dryRun: true
  retentionDays: 14
  exclusions:
    - type: Whitelist
      attributes:
        - key: swabbieResourceOwner
          value:
            - user1@netflix.com
            - user2@netflix.com
```
Whitelist images owned by the specified users and everything else will be excluded.

Exclusion policies provide fine grained control over what gets excluded/whitelisted, including excluding by fields on the resource being scanned.

```
- name: image
  enabled: true
  dryRun: true
  retentionDays: 14
  exclusions:
    - type: Whitelist
      attributes:
        - key: imageId
          value:
            - ami-123
            - ami-132
```

or 

```
- name: image
  enabled: true
  dryRun: true
  retentionDays: 14
  exclusions:
     - type: Literal
        attributes:
         - key: description
            value:
              - pattern:(?!^.*?ancestor_id=)
```
This would exclude images not containing `ancestor_id=`, internally at Netflix this would exclude base images.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
```diff
@@ -0,0 +1,252 @@
...
+  lockingService,
+  retrySupport
+) {
+  override fun remove(markedResource: MarkedResource, workConfiguration: WorkConfiguration) {
+    orcaService.orchestrate(
+      OrchestrationRequest(
+        application = "swabbie",
+        job = listOf(
+          OrcaJob(
+            type = "deleteImage",
```
jeyrschabu: Need a stage to delete images. Also considering a swabbie stage with tasks mapped to a delete action.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
```diff
@@ -0,0 +1,252 @@
...
+  lockingService: Optional<LockingService>,
+  retrySupport: RetrySupport,
+  private val rules: List<Rule<AmazonImage>>,
+  private val imageProvider: ResourceProvider<AmazonImage>,
+  private val instanceProvider: ResourceProvider<AmazonInstance>,
+  private val launchConfigurationProvider: ResourceProvider<AmazonLaunchConfiguration>,
+  private val accountProvider: AccountProvider,
+  private val orcaService: OrcaService
+) : AbstractResourceTypeHandler<AmazonImage>(
+  registry,
```
emjburns: weird spacing, should be indented one?

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
```diff
@@ -0,0 +1,252 @@
...
+  rules,
+  resourceTrackingRepository,
+  exclusionPolicies,
+  resourceOwnerResolver,
+  notifier,
+  applicationEventPublisher,
+  lockingService,
+  retrySupport
+) {
+  override fun remove(markedResource: MarkedResource, workConfiguration: WorkConfiguration) {
```
emjburns: Worth calling this `delete` instead of remove for clarity all the way up? 
Are there any situations where something is not deleted when it is `remove`d?

(Not necessary to change in this PR but just a comment).

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
emjburns: Do you mean that there isn't a stage for that, and that this is just a dummy call for now? If so, would be great to note in a // TODO

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/Rules.kt
```diff
@@ -0,0 +1,51 @@
...
+    val hasSiblings = resource.details.containsKey(HAS_SIBLINGS_IN_OTHER_ACCOUNTS) &&
+      resource.details[HAS_SIBLINGS_IN_OTHER_ACCOUNTS] as Boolean
+
+    if (isReferencedByInstances || isReferencedByLaunchConfigs || hasSiblings) {
+      return Result(null)
+    }
+
+    return Result(
+      Summary(
+        description = "Image is not referenced by an instance, a launch config and has no siblings in other accounts",
```
emjburns: To clarify, an orphan is when  "Image is not referenced by and instance AND image not reference by a launch config AND image has no siblings in other accounts" right? That might make a clearer description.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
jeyrschabu: yeah it is. Verified in the IDE it is on the same line as the constructor members.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
jeyrschabu: Renaming to `delete` is a good idea.
On the following question, I suppose you are referring to when an orca job to delete a resource fails? In that case, the pattern to follow would be to poll Orca for a task until it's completed. That's legit and an action item to tackle before turning `delete` functionality on for real.

jeyrschabu: Will have a PR to address ^^.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
```diff
@@ -0,0 +1,252 @@
...
+        listOf(
+          setUsedByInstancesAsync(images, params),
+          setUsedByLaunchConfigurationsAsync(images, params),
+          setHasSiblingsAsync(images, params)
+        ).forEach { checkingReferencesTask ->
+          try {
+            checkingReferencesTask.await()
+          } finally {
+            if (checkingReferencesTask.isCompletedExceptionally) {
+              val throwable = checkingReferencesTask.getCompletionExceptionOrNull()
```
emjburns: wow this is my favorite name for a boolean ever: `isCompletedExceptionally`

swabbie-edda/src/main/kotlin/com/netflix/spinnaker/swabbie/edda/EddaService.kt
```diff
@@ -43,4 +46,22 @@ interface EddaService {
 
   @GET("/api/v2/aws/autoScalingGroups;_expand")
   fun getAutoScalingGroups(): List<AmazonAutoScalingGroup>
+
+  @GET("/api/v2/aws/images;_expand:(imageId,name,description,state,tags)")
```
emjburns: This whole class works with clouddriver config, right? So that if you don't have edda and are hitting aws directly this does the right thing?

swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandlerTest.kt
```diff
@@ -0,0 +1,501 @@
...
+  @Test
+  fun `should handle work for images`() {
+    Assertions.assertTrue(subject.handles(getWorkConfiguration()))
+  }
+
+  @Test
+  fun `should find image cleanup candidates`() {
+    val params = Parameters(mapOf("account" to "1234", "region" to "us-east-1"))
+    whenever(imageProvider.getAll(params)) doReturn listOf(
+      AmazonImage(
```
emjburns: Would be helpful if you could pull this `whenever(imageProvider.getAll(params)) do return listOf(...)` to the top so that it could be used in every test w/o having to repeat it so much? Nitting just for ease of understanding the tests/making more.

swabbie-edda/src/main/kotlin/com/netflix/spinnaker/swabbie/edda/EddaService.kt
jeyrschabu: Edda is set by default as of now but that will change when we add aws support.
At that point, we should be able to toggle the edda flag and be able to talk to amazon directly.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/Rules.kt
jeyrschabu: To keep it simple initially swabbie will skip those. Next step, we would be more aggressive by checking the validity of that image in that account and in relation to other accounts.

PR #65, opened 2018-07-24
refactor(tagging): Misc refactor with tagging

jeyrschabu:
- move logic out of the web module
- move tagging flag to resource type level


jeyrschabu:
Moving some things around and adding the flag at the resource type level.

PR #67, opened 2018-07-28
feat(exclusions): Adding support for dynamic exclusion suppliers

jeyrschabu:
- added an interface for dynamic exclusions
- added a base ami label supplier to enable excluding them


jeyrschabu:
Allows reading base ami labels from a dynamic property (FP).
These are added as runtime exclusions and would exclude anything matching the name provided in the list.

PR #78, opened 2018-08-08
chore(core): Updated list of resources users were notified for

jeyrschabu:
- using a list of marked resources to pretty print result


jeyrschabu:
@emjburns looks like we forgot to do the same thing for the notify function.

PR #86, opened 2018-08-21
fix(images): synchronously run image classifiers

asher:
- Fixes a concurrency bug in the Image Handler caused by multiple async classifiers updating a non-threadsafe data structure
- Log ImageId, name, and region for each classification


emjburns:
LGTM. I'll follow up in the future on whether it makes sense to add some sort of concurrent map to [Resource](https://github.com/spinnaker/swabbie/blob/master/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt#L79)

PR #87, opened 2018-08-21
feat(core): prefilter resources by maxAge

asher:
- chained image classifiers won't duplicate work

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
@@ -151,12 +151,12 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
       totalResourcesVisitedCounter.set(candidates.size)
 
       preProcessCandidates(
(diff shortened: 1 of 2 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -151,12 +152,12 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
       totalResourcesVisitedCounter.set(candidates.size)
 
       preProcessCandidates(
-        candidates,
+        candidates.filter {
+          it.createTs < Instant.now().toEpochMilli() - TimeUnit.DAYS.toMillis(workConfiguration.maxAge.toLong())
```
emjburns: Did you do this instead of using `shouldExcludeResource()` ([link](https://github.com/spinnaker/swabbie/blob/master/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt#L268)) since you want to filter by just time, not if it's been checked before?

emjburns:
LGTM

PR #88, opened 2018-08-22
chore(*): s/whitelist/allowlist/

robzienert:


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/exclusions/ExclusionPolicy.kt
```diff
@@ -36,8 +36,8 @@ interface ExclusionPolicy {
 
   fun getType(): ExclusionType
 
-  fun notWhitelistedMessage(value: String?, patterns: Set<Any> = emptySet()): String {
-    return "$value not whitelisted. Matched: ${patterns.joinToString(",")}"
+  fun notAllowlistedMessage(value: String?, patterns: Set<Any> = emptySet()): String {
+    return "$value not allowed. Matched: ${patterns.joinToString(",")}"
```
cfieber: `not in allowlist`? `not allowed` feels like it could also be a validation type error

robzienert: Good call.

PR #89, opened 2018-08-23
fix(exclusions): Should resolve owners before running exclusion policies

jeyrschabu:
- this fixes a regression due to refactoring
- need to be able to exclude/whitelist by owner
- to allow exclusion/inclusion by resource owner, that field needs to be computed before applying policies


jeyrschabu:
This is an important fix. Broke exclusion by resource owner.

emjburns:
Oops sorry @jeyrschabu! LGTM

PR #85, opened 2018-08-21
feat(deleteImage): complete image deletions

jeyrschabu:
- IN PROGRESS

PR #91, opened 2018-08-23
feat(deleteImage): Delete image

jeyrschabu:
- implement delete images
- update core to fetch resources eagerly in one call
- refactor delete to be more efficient
- adapt tests


jeyrschabu:
Pulled from https://github.com/spinnaker/swabbie/pull/85.

PR #100, opened 2018-08-28
feat(controller): slim for resources/marked exists as default

emjburns:
Upgrade kotlin version

Add a slim option for marked resource which gives you the identifying info and violations but not the whole resource object. This is the default. Get the full details by going to `https://localhost:8088/resources/marked?expand=true`

Also adding an endpoint to get a single marked resource `https://localhost:8088/resources/marked/{namespace}/{resourceId}`

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
@@ -105,20 +105,60 @@ interface Timestamped {
   val createTs: Long
 }
 
+interface MarkedResourceInterface : Identifiable {
+  val summaries: List<Summary>
+  val namespace: String
+  var projectedDeletionStamp: Long
+  var notificationInfo: NotificationInfo?
+  var markTs: Long?
+  var updateTs: Long?
+  val resourceOwner: String
+}
+
 /**
  * Cleanup candidate decorated with additional metadata
  * 'projectedDeletionStamp' is the scheduled deletion time
  */
 data class MarkedResource(
   val resource: Resource,
-  val summaries: List<Summary>,
-  val namespace: String,
-  var projectedDeletionStamp: Long,
-  var notificationInfo: NotificationInfo? = null,
-  var markTs: Long? = null,
-  var updateTs: Long? = null,
-  val resourceOwner: String = ""
-) : Identifiable by resource
+  override val summaries: List<Summary>,
+  override val namespace: String,
+  override var projectedDeletionStamp: Long,
+  override var notificationInfo: NotificationInfo? = null,
+  override var markTs: Long? = null,
+  override var updateTs: Long? = null,
+  override val resourceOwner: String = ""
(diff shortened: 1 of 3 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
```diff
@@ -105,20 +105,60 @@ interface Timestamped {
   val createTs: Long
 }
 
+interface MarkedResourceInterface : Identifiable {
```
robzienert: So, I'd probably change this to `MarkedResource` and then the current data class of `MarkedResource` to say... `MarkedResourceImpl` or `DetailedMarkedResource` or something.

swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/ResourceController.kt
```diff
@@ -34,8 +32,26 @@ class ResourceController(
   private val applicationEventPublisher: ApplicationEventPublisher,
   private val workConfigurations: List<WorkConfiguration>
   ) {
+
   @RequestMapping(value = ["/marked"], method = [RequestMethod.GET])
-  fun markedResources(): List<MarkedResource> = resourceTrackingRepository.getMarkedResources()
+  fun markedResources(
+    @RequestParam(required = false) expand: Boolean?
```
robzienert: I think you could be able to do something like this instead, which will eliminate the null check...

`@RequestParam(required = false, default = "false") expand: Boolean`

swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/ResourceController.kt
```diff
@@ -34,8 +32,26 @@ class ResourceController(
...
+  ): List<MarkedResourceInterface> {
+    val markedResources = resourceTrackingRepository.getMarkedResources()
+    if (expand == null || expand == false) {
+      return markedResources.map { it.slim() }
+    }
+    return markedResources
```
robzienert: If you wanna be kotlin-y:

```kotlin
return resourceTrackingRepository.getMarkedResources().let { 
    if (expand) it else it.map { it.slim() }
  }
```

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
emjburns: (Discussed offline). I'm gonna leave as is because `MarkedResource` is used everywhere. If it becomes annoying later on I'll change it.

swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/ResourceController.kt
emjburns: woooooow

PR #101, opened 2018-08-28
feat(controller): get marked resources by namespace

emjburns:
Adding some useful apis to drill down into marked resources. this separates by namespace so you can look at all images in an account and region or whatever you'd like


jeyrschabu:
LGTM

PR #55, opened 2018-06-26
feat(scheduledAgent): Should be destroying the scheduler on shutdown

jeyrschabu:
- `@Predestroy` method to shutdown schedule worker

PR #104, opened 2018-09-07
fix(discovery): add discovery activated

emjburns:
Introducing discovery activated like we do in various other Spinnaker services. This allows us to go in/out of service as we are enabled/disabled. This also allows us to run locally with eureka enabled.

Porting destroy function from https://github.com/spinnaker/swabbie/pull/55.


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/agents/ScheduledAgent.kt
```diff
@@ -69,11 +70,19 @@ abstract class ScheduledAgent(
...
       Duration
         .between(it.getLastAgentRun(), clock.instant())
         .toMillis().toDouble()
-    })
+    }
+  }
+
+  @PreDestroy
+  private fun stop() {
+    log.info("Shutting swabbie down")
```
robzienert: I think this log should be changed to something like, `.info("Scheduled action {} is stopping", javaClass.simpleName)`

"Shutting swabbie down" sounds scary and isn't really accurate.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/discovery/DiscoveryActivated.kt
```diff
@@ -0,0 +1,30 @@
...
+ * and stops doing that thing when it goes down.
+ */
+interface DiscoveryActivated : ApplicationListener<RemoteStatusChangedEvent> {
+
+  val onDiscoveryUpCallback: (event: RemoteStatusChangedEvent) -> Unit
+  val onDiscoveryDownCallback: (event: RemoteStatusChangedEvent) -> Unit
+
+  override fun onApplicationEvent(event: RemoteStatusChangedEvent) {
+    if (event.source.status == InstanceInfo.InstanceStatus.UP) {
+      log.info("Instance is {}... {} starting", event.source.status, DiscoveryActivated::class.java.simpleName)
```
robzienert: `getClass().simpleName` or whatever instead so the log messages use the actual class rather than always saying `DiscoveryActivated`

PR #105, opened 2018-09-08
fix(handler): Fixing bug on excluding a candidate

jeyrschabu:
- fixed a stupid but deadly bug on candidate exclusion
- should not be grabbing the same element over in the filter


jeyrschabu:
Had flashback that I wrote a bug :) FYI @emjburns

PR #112, opened 2018-09-24
fix(schedule): time zones, default to UTC for clock

emjburns:
Was incorrectly handling time zones. Now, specifying a UTC clock and adding tests to make sure that we handle time correctly.


swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/agents/ScheduledAgentTest.kt
```diff
@@ -0,0 +1,64 @@
...
+    val zoneId = ZoneId.of("America/Los_Angeles")
+    val dateTime = LocalDateTime.of(2018, Month.SEPTEMBER, 17, 10, 0, 0) // a monday
+    val instant = dateTime.toInstant(ZoneOffset.of("-7"))
+    val clock = Clock.fixed(instant, zoneId)
+
+    assert(ScheduledAgent.timeToWork(Schedule(), clock))
+  }
+
+  @Test
+  fun `monday 8pm am not work time for normal schedule`() {
```
asher: pm am

emjburns: that's the best time

PR #113, opened 2018-10-05
 feat(agents): soft delete images + testing endpoints

emjburns:
This PR is giant, apologies in advance. I'm well aware that each item in the below list should be a separate PR.

1. Add a soft delete agent
This agent must be enabled overall, and for each work configuration.
"soft deleting" is a reversible deletion, implemented for aws images as tagging with a specific value. We're going to use a deny everything security policy to actually prevent an image from launching if it's been soft deleted (todo: surface this in the ui)
This is enabled in the config in each resource type like:
```yaml
...
resourceTypes:
  - name: image
    softDelete:
      enabled: true
      waitingPeriodDays: 2
```

2. Add testing endpoints in ResourceController to mark, soft delete, and delete resources

3. Add "restore" functionality and endpoint (this removes the previously mentioned tag) and sets a resource to opt-out status, in case the resource was erroneously soft deleted

4. Remove channels, instead anything that needs to be monitored (because it's an orca task) is monitored through OrcaTaskMonitoringAgent, and events are emitted when that task is done. These are processed in ResourceStateManager and ResourceTrackingManager

5. Added endpoints to see resources by state (what had an orca task about it but failed? what was soft deleted?)

6. Added a testing rule so that a resource is always up for cleaning. Must be enabled in config like ``` yaml
swabbie:
  testing:
    alwaysCleanRuleConfig: 
      enabled: true
      resourceIds:
        - ami-blahblah
```

### CODE REVIEW

File: swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt [hunk 1]
```diff
@@ -84,101 +86,90 @@ class AmazonImageHandler(
...
-            send(it)
-          }
-        }
-        taskStatus.isFailure() -> throw IncompleteOrFailedDeletionException("Failed to delete $markedResources in orca task $taskId")
-        taskStatus.isIncomplete() -> {
-          val deleteChannel = monitorDelete(taskId, markedResources, workConfiguration)
-          deleteChannel.consumeEach { send(it) }
-        }
-      }
+      taskTrackingRepository.add(
```
emjburns: The OrcaTaskMonitoringAgent now periodically monitors these tasks, and emits spring events when they're finished.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -117,15 +141,72 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
...
         callback.invoke()
-      })
+      }
     } else {
       log.warn("***Locking not ENABLED***")
       callback.invoke()
     }
   }
 
+  override fun markResource(resourceId: String, onDemandMarkData: OnDemandMarkData, workConfiguration: WorkConfiguration) {
```
emjburns: This is on-demand mark functionality to make it easier to test.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -305,9 +402,11 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
     optedOutResourceStates: List<ResourceState>,
     action: Action
   ): Boolean {
+    if (resource.getViolations().any { summary -> summary.ruleName == "AlwaysCleanUpRule" }) return false
```
emjburns: Makes sure that if the `AlwaysCleanUpRule` is enabled (turned on in config file) that a resource isn't excluded from cleaning based on actual reasons (like it's in use). This is only for testing!

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/repositories/InMemoryTaskTrackingRepository.kt
```diff
@@ -0,0 +1,86 @@
...
+ *  * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.swabbie.repositories
+
+import java.time.Clock
+import java.util.concurrent.TimeUnit
+
+class InMemoryTaskTrackingRepository(
```
emjburns: For testing but probably not necessary.

swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt
```diff
@@ -0,0 +1,190 @@
...
+
+  private fun monitorOrcaTasks() {
+    initialize()
+
+    val inProgressTasks = taskTrackingRepository.getInProgress()
+    if (inProgressTasks.isEmpty()) {
+      log.debug("No active orca tasks to monitor from ${javaClass.simpleName}.")
+    }
+
+    inProgressTasks
```
emjburns: When there are orca tasks to monitor, this agent monitors them (checks status) every 10 seconds (configurable), and emits the appropriate event when they complete. This is the substitute for `ReceiveChannel`s. 

There should be locking around reading from redis in - I'll add that in.

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisTaskTrackingRepository.kt
```diff
@@ -0,0 +1,139 @@
...
+import com.netflix.spinnaker.swabbie.repositories.TaskTrackingRepository
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.stereotype.Component
+import java.time.Clock
+import java.util.concurrent.TimeUnit
+
+@Component
+class RedisTaskTrackingRepository(
```
emjburns: The redis impl of the repository tracking all the orca task. Am I doing anything in here that is inherently not safe for multiple instances?

swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt
emjburns: Added locking.

swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandlerTest.kt
```diff
@@ -364,6 +351,7 @@ object AmazonImageHandlerTest {
   @Test
   fun `should delete images`() {
     val fifteenDaysAgo = System.currentTimeMillis() - 15 * 24 * 60 * 60 * 1000L
+    val thirteenDaysAgo = System.currentTimeMillis() - 13 * 24 * 60 * 60 * 1000L
```
robzienert: Using real times in tests is asking for trouble. Should use a `Clock` with a fixed time, then use instants with calculated times for these sorta things.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -272,9 +355,23 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
     }
   }
 
-  private fun deletionTimestamp(workConfiguration: WorkConfiguration): Long =
+  // todo eb clean up name
+  // now + retention period
```
robzienert: Do these need to be cleaned up prior to merge?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
robzienert: These rule names should be constants.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceTrackingManager.kt
```diff
@@ -0,0 +1,63 @@
...
+@Component
+class ResourceTrackingManager(
+  private val resourceTrackingRepository: ResourceTrackingRepository,
+  private val clock: Clock,
+  private val registry: Registry
+) : MetricsSupport(registry) {
+
+  private val log: Logger = LoggerFactory.getLogger(javaClass)
+
+
```
robzienert: extra space

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceTrackingManager.kt
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * Other event types are handled in {@link AbstractResourceTypeHandler}.
+   * These events are emitted after an orca task is completed in
+   * {@link OrcaTaskMonitoringAgent}
```
robzienert: Don't need to use javadoc style in kotlin: Just do `[OrcaTaskMonitoringAgent]`.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/AlwaysCleanUpRule.kt
```diff
@@ -0,0 +1,48 @@
...
+package com.netflix.spinnaker.swabbie.model
+
+import com.netflix.spinnaker.config.SwabbieProperties
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression
+import org.springframework.stereotype.Component
+
+/**
+ * Testing rule that is invalid on everything.
+ */
+@ConditionalOnExpression("\${swabbie.testing.alwaysCleanRuleConfigConfig.enabled:false}")
```
robzienert: `ConfigConfig`?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/OnDemandMarkData.kt
```diff
@@ -0,0 +1,16 @@
...
+
+data class OnDemandMarkData(
+  var projectedSoftDeletionStamp: Long,
+  var projectedDeletionStamp: Long,
+  var markTs: Long? = null,
+  var resourceOwner: String = "swabbie@spinnaker.io",
+  var notificationInfo: NotificationInfo? = NotificationInfo(
+    recipient = resourceOwner,
+    notificationType = Notifier.NotificationType.EMAIL.name,
+    notificationStamp = 1538589958608,
```
robzienert: Is a static timestamp desired?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/WorkConfiguration.kt
```diff
@@ -40,3 +41,8 @@ class EmptyNotificationConfiguration : NotificationConfiguration(
   types = mutableListOf(),
   optOutBaseUrl = ""
 )
+
+data class SoftDelete(
+  var enabled: Boolean = false, // controls if soft deleting is enabled for this work configuration
+  var waitingPeriodDays: Int = 2 // How many days swabbie will wait after soft deleting a resource before deleting it
```
robzienert: Prefer this documentation form for data classes & constructors, as it'll actually get picked up by doc generators:

```
/**
 * @param enabled controls if soft deleting is...
 */
data class SoftDelete( ... )
```

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/repositories/InMemoryTaskTrackingRepository.kt
robzienert: For all repositories like this, I'd recommend adding a logger and an init that reports what repository is being used.

```
init {
  log.info("Using ${javaClass.simpleName}")
}
```

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/repositories/ResourceStateRepository.kt
```diff
@@ -14,12 +14,14 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.swabbie
+package com.netflix.spinnaker.swabbie.repositories
```
robzienert: Singular plz.

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/ResourceTypeHandlerTest.kt
```diff
@@ -578,17 +601,50 @@ object ResourceTypeHandlerTest {
     override fun deleteResources(
       markedResources: List<MarkedResource>,
       workConfiguration: WorkConfiguration
-    ): ReceiveChannel<MarkedResource> = produce {
+    ) {
+      markedResources.forEach { m ->
+        simulatedCandidates
+          ?.removeIf { r -> m.resourceId == r.resourceId }.also {
+            //todo eb: add to task tracking repo?
+            if (it != null && it) {
```
robzienert: This is a strange conditional? `it` is a `Boolean`?

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/events/ResourceTrackingManagerTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    clock = clock,
+    registry = registry
+  )
+
+  @AfterEach
+  fun cleanup() {
+    reset(resourceTrackingRepository)
+  }
+
+
```
robzienert: space

swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt
```diff
@@ -0,0 +1,214 @@
...
+        .replace(":", ".")
+        .toLowerCase()
+      val lockOptions = LockManager.LockOptions()
+        .withLockName(normalizedLockName)
+        .withMaximumLockDuration(lockingService.get().swabbieMaxLockDuration)
+      lockingService.get().acquireLock(lockOptions) {
+        callback.invoke()
+      }
+    } else {
+      log.warn("***Locking not ENABLED***")
```
robzienert: This log needs more context; like the lock name.

swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt
```diff
@@ -0,0 +1,214 @@
...
+      log.debug("No active orca tasks to monitor from ${javaClass.simpleName}.")
+    }
+
+    inProgressTasks
+      .forEach { taskId ->
+        val response = getTask(taskId)
+        if (response.status.isComplete()) {
+          val taskInfo = taskTrackingRepository.getTaskDetail(taskId)
+          if (taskInfo == null) {
+            log.error("Detail for task $taskId was not found in repository. Something went wrong!")
```
robzienert: This isn't a particularly helpful log message. If something went wrong, we should do what we can to provide good context into what actually went wrong. Think about 6 months, a year from now and we see this log message--will future Emily hate past Emily when triaging a production issue?

swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt
```diff
@@ -0,0 +1,214 @@
...
+          }
+
+          when {
+            response.status.isSuccess() -> {
+              log.debug("Orca task $taskId succeeded. Task complete info: $taskInfo")
+              publishEvent(taskInfo)
+              taskTrackingRepository.setSucceeded(taskId)
+            }
+            response.status.isFailure() -> {
+              log.error("Orca task $taskId did not complete. Status: ${response.status}. Task complete info: $taskInfo")
```
robzienert: We need to go through all of swabbie and get it logging structured logs, but it's not too late for new code. Use `StructuredArguments.kv` for things like taskId, status, etc so we can index and search through the logs better.

swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt
```diff
@@ -0,0 +1,214 @@
...
+              taskTrackingRepository.setFailed(taskId)
+              taskInfo.markedResources
+                .forEach { markedResource ->
+                  applicationEventPublisher.publishEvent(OrcaTaskFailureEvent(taskInfo.action, markedResource, taskInfo.workConfiguration))
```
robzienert: Mind column 120.

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceTrackingRepository.kt
```diff
@@ -34,76 +35,107 @@ class RedisResourceTrackingRepository(
...
+    return try {
+      redisClientDelegate.withCommandsClient<String> { client ->
         client.hget(SINGLE_RESOURCES_KEY, key)
-      }.let { json ->
-          objectMapper.readValue(json)
-        }
+      }.let { objectMapper.readValue(it) }
+    } catch (e: IllegalStateException) {
+      // resource does not exist
+      null
```
robzienert: Mmm... this doesn't sound right. What is throwing ISE? The redis client or the object mapper? If it's the object mapper, the resource exists, but it's malformed.

swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt
```diff
@@ -0,0 +1,214 @@
...
+  private val swabbieProperties: SwabbieProperties,
+  private val taskTrackingRepository: TaskTrackingRepository,
+  private val orcaService: OrcaService,
+  private val applicationEventPublisher: ApplicationEventPublisher,
+  private val retrySupport: RetrySupport,
+  private val lockingService: Optional<LockingService>
+) : DiscoveryActivated, MetricsSupport(registry) {
+
+  private val log: Logger = LoggerFactory.getLogger(javaClass)
+  private val worker: Scheduler.Worker = Schedulers.io().createWorker()
```
robzienert: Let's avoid the propagation of Rx: We're trying to get rid of it from all of our projects. Prefer the built-in Java ExecutorService.

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceTrackingRepository.kt
```diff
@@ -34,76 +35,107 @@ class RedisResourceTrackingRepository(
...
-          }
+  override fun getIdsOfMarkedResourcesToSoftDelete(): Set<String> {
+    return getAllIds(SOFT_DELETE_KEY, false)
+  }
+
+  override fun getIdsOfMarkedResourcesToDelete(): Set<String> {
+    return getAllIds(DELETE_KEY, false)
+  }
+
+  private fun getAllIds(key: String, includeAll: Boolean): Set<String> {
```
robzienert: `includeAll` doesn't really do a good job of describing what the behavior is when false, especially when the method is called `getAllIds`, I'd expect it would always provide all ids, regardless of an `includeAll` being true or false.

Perhaps `getIds(key: String, includeFutureIds: Boolean)` or something to that effect?

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceTrackingRepository.kt
```diff
@@ -34,76 +35,107 @@ class RedisResourceTrackingRepository(
...
-  override fun upsert(markedResource: MarkedResource, score: Long) {
-    "${markedResource.namespace}:${markedResource.resourceId}".let { id ->
-      markedResource.apply {
-        markTs = if (markTs != null) markTs else Instant.now(clock).toEpochMilli()
-        updateTs = if (markTs != null) Instant.now(clock).toEpochMilli() else null
+  private fun hydrateMarkedResources(resourseIds: Set<String>): List<MarkedResource> {
+    if (resourseIds.isEmpty()) return emptyList()
+    return redisClientDelegate.run {
+      this.withCommandsClient<Set<String>> { client ->
+        client.hmget(SINGLE_RESOURCES_KEY, *resourseIds.map { it }.toTypedArray()).toSet()
```
robzienert: `*resourceIds.toTypedArray()` will suffice.

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceTrackingRepository.kt
```diff
@@ -34,76 +35,107 @@ class RedisResourceTrackingRepository(
...
-          client.zrem(ALL_RESOURCES_KEY, id)
-          client.hdel(SINGLE_RESOURCES_KEY, id)
-        }
-      }
+    val id = markedResource.uniqueId()
+    redisClientDelegate.withCommandsClient { client ->
+      client.zrem(DELETE_KEY, id)
+      client.zrem(SOFT_DELETE_KEY, id)
+      client.hdel(SINGLE_RESOURCES_KEY, id)
     }
```
robzienert: Would recommend adding retries where possible around redis operations. This isn't done in many other places in Spinnaker, but for the new persistence backends I've been working on, I've been ensuring retries are put around all interactions so transient network failures don't cause as many application errors.

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisTaskTrackingRepository.kt
```diff
@@ -0,0 +1,139 @@
...
+      client.hset(TASK_STATUS_KEY, taskId, TaskState.IN_PROGRESS.toString())
+    }
+  }
+
+  override fun isInProgress(taskId: String): Boolean {
+    redisClientDelegate.run {
+      return this.withCommandsClient<Boolean> { client ->
+        client.hget(TASK_STATUS_KEY, taskId) == TaskState.IN_PROGRESS.toString()
+      }
+    }
```
robzienert: The `run` call seems superfluous.

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisTaskTrackingRepository.kt
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  private fun getAllBefore(daysToGoBack: Int): Set<String> {
+    return getAll(SUBMITTED_TASKS_KEY)
+      .filterValues { value ->
+        objectMapper.readValue(value, TaskCompleteEventInfo::class.java).let { taskInfo ->
+          val xDaysAgo = clock.instant().minusMillis(TimeUnit.DAYS.toMillis(daysToGoBack.toLong())).toEpochMilli()
+          taskInfo.submittedTimeMillis!! < xDaysAgo
```
robzienert: `!!` is usually a code smell. Are you sure you want to hard-crash the process if the submitted time is null?

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisTaskTrackingRepository.kt
```diff
@@ -0,0 +1,139 @@
...
+
+  //todo eb: this actually cleans up running tasks as well
+  override fun cleanUpFinishedTasks(daysToLeave: Int) {
+    val allBefore: Set<String> = getAllBefore(daysToLeave)
+    log.debug("Cleaning ${allBefore.size} tasks")
+    allBefore.forEach { taskId ->
+      redisClientDelegate.withCommandsClient { client ->
+        client.hdel(TASK_STATUS_KEY, taskId)
+        client.hdel(SUBMITTED_TASKS_KEY, taskId)
+      }
```
robzienert: We can save some network calls here...

```kotlin
redisClientDelegate.withCommandsClient { client ->
  client.hdel(TASK_STATUS_KEY, *allBefore)
  client.hdel(SUBMITTED_TASKS_KEY, *allBefore)
}
```

swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/ResourceController.kt
```diff
@@ -71,13 +70,33 @@ class ResourceController(
   }
 
   @RequestMapping(value = ["/canDelete"], method = [RequestMethod.GET])
-  fun markedResourcesReadyForDeletion(): List<MarkedResource> = resourceTrackingRepository.getMarkedResourcesToDelete()
+  fun markedResourcesReadyForDeletion(): List<MarkedResource> =
+    resourceTrackingRepository.getMarkedResourcesToDelete()
+
+  @RequestMapping(value = ["/states/{status}"], method = [RequestMethod.GET])
+  fun getResourcesByState(
+    @PathVariable status: String
```
robzienert: Did you try making the `status` value the `Action` type? Spring is smart enough to handle enums on its own.

robzienert: Not totally sure what `"FAILED"` is about... seems we're conflating status and actions as a single value?

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroupHandler.kt
```diff
@@ -63,10 +66,21 @@ class AmazonAutoScalingGroupHandler(
...
   retrySupport
 ) {
+
+  override fun softDeleteResources(markedResources: List<MarkedResource>, workConfiguration: WorkConfiguration) {
+    TODO("not implemented")
+  }
+
+  override fun restoreResources(markedResources: List<MarkedResource>, workConfiguration: WorkConfiguration) {
+    TODO("not implemented")
+  }
```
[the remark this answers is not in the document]
emjburns: I'm not implementing functionality for auto scaling groups/ security groups/ load balancers until we straighten out the functionality for images.

swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/ResourceController.kt
```diff
@@ -121,73 +140,99 @@ class ResourceController(
...
+  fun delete(
+    @PathVariable resourceId: String,
+    @PathVariable namespace: String
+  ) {
+    val workConfiguration = findWorkConfiguration(SwabbieNamespace.namespaceParser(namespace))
+    val handler = resourceTypeHandlers.find { handler ->
+      handler.handles(workConfiguration)
+    } ?: throw NotFoundException("No handlers for $namespace")
+
+    handler.deleteResource(resourceId, workConfiguration)
```
robzienert: This delete & softdelete methods don't feel right.

Why not have a single endpoint:

```
@RequestMapping(value = ["/state/{namespace}/{resourceId}"], method = [RequestMethod.DELETE])
fun delete(
  @PathVariable resourceId: String,
  @PathVariable namespace: String,
  @QueryParam(required = true) soft: Boolean
)
```

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
emjburns: Oops. good catch

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
```diff
@@ -48,14 +50,16 @@ class AmazonImageHandler(
   exclusionPolicies: List<ResourceExclusionPolicy>,
   applicationEventPublisher: ApplicationEventPublisher,
   lockingService: Optional<LockingService>,
-  private val retrySupport: RetrySupport,
+  retrySupport: RetrySupport,
```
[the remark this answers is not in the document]
emjburns: In Kotlin, doesn't using this syntax mean it's not a variable on the class but just used to initialize the object?

swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandlerTest.kt
emjburns: Arg! Missed that one. Fixing.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
emjburns: Good call. I'll remove the todo. Already decided they were fine.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceTrackingManager.kt
emjburns: Fixing everywhere

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/AlwaysCleanUpRule.kt
emjburns: Fixed. Intellij refactor gone wrong.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/OnDemandMarkData.kt
emjburns: I'm not sure? This is for testing, so providing these default values let you submit a marked resource w/o specifying all notification info lets the resource actually be cleaned (resource won't be cleaned if it hasn't been notified first). The time doesn't particularly matter. You can override it in your payload body when you submit via the api. I'd say static is fine for now.

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/ResourceTypeHandlerTest.kt
emjburns: yes. Fixed the naming so it's more obvious.

swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt
emjburns: Made it a bit better... 
`log.error("TaskDetail not found in tracking repository for $taskId. Unable to fire completion event for task.")` 
I don't know when this case would happen. Redis problems? The wrong tasks being cleaned out of the task repository? So I can't give a more helpful message.

swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt
emjburns: I'll use this for new things and clean up as I go. Not gonna add more sweeping changes to this PR

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceTrackingRepository.kt
emjburns: Oh I think I just need the handy `?` because the object mapper was complaining the value was null. Not fully used to the awesomeness of Kotlin.

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceTrackingRepository.kt
emjburns: @robzienert happy to. Do you recommend using RetrySupport?

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisTaskTrackingRepository.kt
emjburns: We set this value when the object is added to the repository (see L52). Let me think about handling it in a different way..

swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/ResourceController.kt
emjburns: Did not try. 

With FAILED - trying to record when an action has failed on the orca side with good visibility, but there is no opportunity for a message field on the Status object I'm leaning towards adding an optional message on the status object so that it can be in state "FAILED" with a message about what action failed. Any thoughts on how to not make this feel weird?

swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/ResourceController.kt
emjburns: Good idea will change

swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/ResourceController.kt
emjburns: Will address this in a different PR.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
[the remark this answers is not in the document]
robzienert: Yes nvm.

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisTaskTrackingRepository.kt
emjburns: Fixed.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/OnDemandMarkData.kt
robzienert: Even if the intention today is for testing-only purposes, we can't be sure that its scope will remain a testing-only function in the future. I'd say it shouldn't have any default value assigned and should be explicitly set by the caller. These sort of assumptions have a higher chance of creating bugs some years down the line.

swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt
robzienert: Hmm, based on the current implementation of the repository, a Redis problem would throw an exception and it wouldn't even get to this conditional. The updated log message is improved!

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/OnDemandMarkData.kt
emjburns: Great point. Removed the timestamp but kept notification for easy testing.

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceTrackingRepository.kt
emjburns: Ok. Will fix in another PR.

PR #123, opened 2018-10-23
feat(mark): removing mark of deleted resources, comments

emjburns:
- adding javadocs
- adding (and logging) a reason for why something is unmarked
- removing resources that no longer exist from the marked database before we run the mark cycle. This makes it easier to view swabbie's list of marked resources, since after a resource is deleted not using swabbie, swabbie will update the state it has.
- add an allow list test for an example


asher:
LGTM

PR #124, opened 2018-10-24
feat(agents): cache of global aws launch configs

asher:
- provides efficient per-region (ami -> set(launchConfigs)) lookups
- cache is built from all eddas, to include accounts not configured in clouddriver
- fix edda lookups by mapping accounts to correct environment


swabbie-edda/src/main/kotlin/com/netflix/spinnaker/swabbie/edda/providers/EddaLaunchConfigurationCacheProvider.kt
```diff
@@ -0,0 +1,105 @@
...
+import com.netflix.spinnaker.swabbie.*
+import com.netflix.spinnaker.swabbie.aws.launchconfigurations.AmazonLaunchConfiguration
+import com.netflix.spinnaker.swabbie.model.WorkConfiguration
+import org.springframework.context.annotation.Configuration
+import org.springframework.context.annotation.Lazy
+import java.lang.IllegalArgumentException
+import java.time.Clock
+
+@Configuration
+open class EddaLaunchConfigurationCacheProvider(
```
asher: `EddaLaunchConfigurationCacheProvider` provides:
* A cache of all `AmazonLaunchConfigs` for all accounts by region
* A cached hashmap keyed on AMI's actively referenced by a launch config in any account (by region), where values are a set of `AmazonLaunchConfig` currently referencing the key.
* Updated every N minutes (currently 15min based on the `Cacheable` default)
* Provides a `lastUpdated` attribute

Example use:
`AmazonImageHandler` could inject` launchConfigurationCache: CachedViewProvider<AmazonLaunchConfigurationCache>` and call: `val amisUsedByLaunchConfigs = launchConfigurationCache.getRefdAmisForRegion(region)`, then filter candidates where  `amisUsedByLaunchConfigs.contains(amiId)`. If `launchConfigurationCache.getLastUpdated()` is older than an expected value (i.e. now-1h), abort evaluation.

swabbie-edda/src/main/kotlin/com/netflix/spinnaker/swabbie/edda/providers/EddaApiSupport.kt
```diff
@@ -29,9 +29,9 @@ open class EddaApiSupport(
 ) {
   private val log: Logger = LoggerFactory.getLogger(javaClass)
   val eddaFailureCountId: Id = registry.createId("swabbie.edda.failures")
-  fun withEddaClient(region: String, accountId: String): EddaService? {
+  fun withEddaClient(region: String, accountId: String, environment: String): EddaService? {
```
asher: This addresses an issue caused by a difference in the way clouddriver and edda represent AWS accounts partitioned into >1 environment. Environments are treated as distinct accounts by both clouddriver and edda, but they share the same AWS account id.

If AWS account spinnaker (id: 1234) was partitioned into test and prod envs, clouddriver would expose them as accounts spinnaker and spinnakertest. Edda would index them under instances edda-1234.test and edda-1234.prod. Each edda instance would only index their environment, not the full 1234 account.

Swabbie made an assumption that AWS account id's were unique; all resource providers fetched a single edda client looked up by account id and region. This could result in the incorrect edda instance being queried for partitioned accounts. Its also possible that some environments would have been excluded from edda lookups, i.e. both spinnaker and spinnakertest accounts could both potentially contain spinnakertest resources.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/securitygroups/AmazonSecurityGroupHandler.kt
```diff
@@ -135,15 +136,15 @@ class AmazonSecurityGroupHandler(
     TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
   }
 
-  override fun handles(workConfiguration: WorkConfiguration): Boolean
-    = workConfiguration.resourceType == SECURITY_GROUP && workConfiguration.cloudProvider == AWS && !rules.isEmpty()
+  override fun handles(workConfiguration: WorkConfiguration): Boolean = workConfiguration.resourceType == SECURITY_GROUP && workConfiguration.cloudProvider == AWS && !rules.isEmpty()
```
emjburns: line limit

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Account.kt
```diff
@@ -43,7 +44,8 @@ data class SpinnakerAccount(
   override val type: String,
   override val name: String,
   override val edda: String?,
-  override val regions: List<Region>?
+  override val regions: List<Region>?,
+  override val environment: String?
```
emjburns: When would environment be null? When this is not Edda backed, and instead a generic account?

swabbie-edda/src/main/kotlin/com/netflix/spinnaker/swabbie/edda/providers/EddaApiSupport.kt
emjburns: Great find.

swabbie-edda/src/main/kotlin/com/netflix/spinnaker/swabbie/edda/providers/EddaLaunchConfigurationCacheProvider.kt
```diff
@@ -0,0 +1,105 @@
...
+  fun load(): Set<AmazonLaunchConfigurationCache> {
+    val configsByRegion: MutableMap<String, Set<AmazonLaunchConfiguration>> = mutableMapOf()
+    val refdAmisByRegion: MutableMap<String, MutableMap<String, MutableSet<AmazonLaunchConfiguration>>> = mutableMapOf()
+
+    val regions = workConfigurations.asSequence()
+      .map { it.location }
+      .toSet()
+
+    regions.forEach { region: String ->
+      val configs: Set<AmazonLaunchConfiguration> = eddaApiClients
```
emjburns: can we call this launchConfigs for clarity?

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroupHandler.kt
```diff
@@ -120,12 +119,15 @@ class AmazonAutoScalingGroupHandler(
...
-    = workConfiguration.resourceType == SERVER_GROUP && workConfiguration.cloudProvider == AWS && !rules.isEmpty()
+  override fun handles(workConfiguration: WorkConfiguration): Boolean = workConfiguration.resourceType == SERVER_GROUP && workConfiguration.cloudProvider == AWS && !rules.isEmpty()
 
   override fun getCandidates(workConfiguration: WorkConfiguration): List<AmazonAutoScalingGroup>? {
     val params = Parameters(
-      mapOf("account" to workConfiguration.account.accountId!!, "region" to workConfiguration.location)
+      mapOf(
+        "account" to workConfiguration.account.accountId!!,
+        "region" to workConfiguration.location,
+        "environment" to workConfiguration.account.environment!!
```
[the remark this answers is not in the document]
emjburns: Seems like the core of this is around the parameters objects and maybe generic accounts vs aws edda-enabled accounts. I agree, this can and should be better.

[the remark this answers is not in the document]
asher: I'd like to see the Parameters(Map<String,Any>) class go away in favor of typed constructors. For now, I'm going to remove nullibility from the Account.environment property I added.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Account.kt
asher: It is my understanding that account environment is primarily a netflix construct that isn't used universally in OSS installs, but not certain. I'm going to remove nullability here for now, in response to Rob's feedback.

swabbie-edda/src/main/kotlin/com/netflix/spinnaker/config/EddaConfiguration.kt
```diff
@@ -62,15 +66,32 @@ open class EddaConfiguration {
         )
       }
     }.flatten()
+
+    val endpointEddas: List<EddaApiClient> = endpointProvider.getEndpoints().filter { endpoint ->
```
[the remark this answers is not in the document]
asher: this is top-notch formatting feedback, appreciated

PR #126, opened 2018-10-24
feat(lastSeen): adding last seen info to markedResource

emjburns:
Making sure that if we've never seen a resource in use we don't mark it as SEEN_IN_USE. That way, when Swabbie starts up, it can both track usage and delete resources that we've never seen in use (like the old amis)

Storing the last seen info with marked resources, so that we can send the information in email notifications to make the more useful.

Sending the resources to echo as their `.slim()` version. We actually don't even need that much data, but I'll pair it down later. We're only rendering the good bits in the email.

### DIFF
--- a/swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceUseTrackingRepository.kt
+++ b/swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceUseTrackingRepository.kt
@@ -83,6 +83,20 @@ class RedisResourceUseTrackingRepository(
     return hydrateLastSeen(keys)
   }
 
+  override fun getUsed(): Set<String> {
+    return redisClientDelegate.withCommandsClient<Set<String>> { client ->
+      client.zrangeByScore(READY_FOR_CLEANING, clock.instant().toEpochMilli().toDouble(), Double.MAX_VALUE)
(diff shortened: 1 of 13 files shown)


swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceUseTrackingRepository.kt
```diff
@@ -83,6 +83,24 @@ class RedisResourceUseTrackingRepository(
...
+    return redisClientDelegate.withCommandsClient<Set<String>> { client ->
+      client.zrangeByScore(READY_FOR_CLEANING, clock.instant().toEpochMilli().toDouble(), Double.MAX_VALUE)
+    }
+  }
+
+  override fun getLastSeenInfo(resourceIdentifier: String): LastSeenInfo? {
+    return redisClientDelegate.withCommandsClient<String> { client ->
+      try {
+        client.hget(LAST_SEEN, resourceIdentifier)
+      } catch (e: NullPointerException) {
```
robzienert: Does `hget` actually throw an NPE?

emjburns: No. Good call. Fixing!

PR #128, opened 2018-10-25
chore(params): map to actual object

emjburns:
Using an object, with fields and types and such.

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/ResourceProvider.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/ResourceProvider.kt
@@ -32,4 +32,9 @@ interface ResourceProvider<out T : Resource> {
   fun getOne(params: Parameters): T?
 }
 
-class Parameters(val map: Map<String, Any>) : HashMap<String, Any>(map)
+data class Parameters(
+  val region: String,
+  val account: String,
+  val environment: String,
+  val id: String = ""
+)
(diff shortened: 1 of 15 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/ResourceProvider.kt
```diff
@@ -32,4 +32,9 @@ interface ResourceProvider<out T : Resource> {
   fun getOne(params: Parameters): T?
 }
 
-class Parameters(val map: Map<String, Any>) : HashMap<String, Any>(map)
+data class Parameters(
+  val region: String,
+  val account: String,
+  val environment: String,
+  val id: String = ""
```
asher: It looks like the Parameters map currently also holds `"autoScalingGroupName": String` and `"imageId": String` in places.

emjburns: Ya. I've updated all of those to be passed as "id" (since that's essentially what it is - a specific id to query for)

emjburns: I don't think there's a need to capture what the field originally was.

asher: I should have read the PR more closely :)

PR #129, opened 2018-10-26
Better in use checking for images

emjburns:
Creates singleton cache, useful for having an object that you want to refresh.
Use this singleton cache in the edda caches (and refactor those to not be `@Lazy`).
Update the `checkReferences(..)` logic in `AmazonImageHandler` to check if an image is used in any launch config in any account, then check if the image is used by any images in any account. 
Rip out the siblings check, as it is no longer needed because we are checking the use in all accounts.

### DIFF
--- a/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
+++ b/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
@@ -21,9 +21,10 @@ import com.netflix.spinnaker.config.SwabbieProperties
 import com.netflix.spinnaker.kork.core.RetrySupport
 import com.netflix.spinnaker.moniker.frigga.FriggaReflectiveNamer
 import com.netflix.spinnaker.swabbie.*
-import com.netflix.spinnaker.swabbie.aws.instances.AmazonInstance
-import com.netflix.spinnaker.swabbie.aws.launchconfigurations.AmazonLaunchConfiguration
+import com.netflix.spinnaker.swabbie.aws.edda.providers.AmazonImagesUsedByInstancesCache
+import com.netflix.spinnaker.swabbie.aws.edda.providers.AmazonLaunchConfigurationCache
 import com.netflix.spinnaker.swabbie.events.Action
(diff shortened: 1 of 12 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/Cacheable.kt
```diff
@@ -58,3 +62,27 @@ open class InMemoryCache<out T : Cacheable>(
 
   val log: Logger = LoggerFactory.getLogger(javaClass)
 }
+
+open class InMemorySingletonCache<out T : Cacheable>(
```
asher: Not at all necessary, but Cacheable inherits from Named which adds the name field requirement (though it is nullable.) Named resources make sense when always caching a set of resources, but it doesn't make sense for the singleton case. It'd be nice to remove the name="default" designation from the singleton caches.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
```diff
@@ -241,119 +240,66 @@ class AmazonImageHandler(
...
-          image
-        ) { usedByResource ->
-          image.set(USED_BY_INSTANCES, true)
-          resourceUseTrackingRepository.recordUse(image.resourceId, usedByResource)
-          log.debug("Image {} ({}) in {} is USED_BY_INSTANCES", image.imageId, image.name, params.region)
-        }
-      }
+    if (launchConfigurationCache.get().getLastUpdated()
+        < clock.instant().toEpochMilli().minus(Duration.ofHours(1).toMillis())) {
+      throw StaleCacheException("Amazon launch configuration cache over 1 hour old, aborting.")
```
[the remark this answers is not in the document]
emjburns: Good idea. Will update!

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/Cacheable.kt
emjburns: I'll do that later - want to properly think through a refactor.

PR #130, opened 2018-10-29
refactor(tracking): refactor last_seen redis index

asher:


swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceUseTrackingRepository.kt
```diff
@@ -62,30 +62,34 @@ class RedisResourceUseTrackingRepository(
...
-          LastSeenInfo(resourceIdentifier, usedByResourceIdentifier, clock.instant().toEpochMilli()))
+          LastSeenInfo(resourceIdentifier, usedByResourceIdentifier, now)
+        )
       )
       // add to sorted set by score
-      // score = day when resource will be ready to delete
+      // score = last_seen timestamp
       // if resource is seen in next X days, resource will be updated
-      client.zadd(READY_FOR_CLEANING, plusXdays(outOfUseThresholdDays).toDouble(), resourceIdentifier)
+      tx.zadd(LAST_SEEN_INDEX, now.toDouble(), resourceIdentifier)
```
asher: Indexing on last_seen instead of a precomputed cleanup date (if a resource isn't seen again) allows us to adjust when we consider no-longer-seen resources as ready to delete vs. only applying to currently active resources.

PR #140, opened 2018-11-06
chore(build): gradle 4.10 / spinnaker-gradle-project 5.2.1

cfieber:


build.gradle
```diff
@@ -20,23 +20,24 @@ buildscript {
     maven { url "https://plugins.gradle.org/m2/" }
   }
   dependencies {
-    classpath "com.netflix.spinnaker.gradle:spinnaker-gradle-project:3.16.0"
-    classpath "org.springframework.boot:spring-boot-gradle-plugin:1.5.9.RELEASE"
+    classpath "com.netflix.spinnaker.gradle:spinnaker-dev-plugin:5.2.1"
     classpath "org.junit.platform:junit-platform-gradle-plugin:1.0.0"
     classpath "com.netflix.nebula:nebula-kotlin-plugin:$kotlinVersion"
     classpath "org.jetbrains.dokka:dokka-gradle-plugin:0.9.15"
-    classpath "com.netflix.nebula:gradle-dependency-lock-plugin:4.9.5"
```
cfieber: I think somebody had intentions of using dependency locking that were never actually realized

I took this out of here for now, but the other change that would be good to make would be to go through the gradle files and get rid of all the version range style specifiers (`1.9.+` etc) since right now this build is not stable and repeatable

build.gradle
```diff
@@ -20,23 +20,24 @@ buildscript {
...
-  apply plugin: "spinnaker.project"
-  apply plugin: "nebula.dependency-lock"
+  apply plugin: "spinnaker.base-project"
+  apply plugin: "java"
 
   group = "com.netflix.spinnaker.swabbie"
 
   ext {
-    spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '1.6.0'
+    spinnakerDependenciesVersion = '1.6.0'
```
cfieber: slight reformatting to appease the autobump dependencies gods

build.gradle
emjburns: @cfieber will do. I've also been slowly moving everything to the spinnaker-dependencies version instead of specifying here, where applicable.

PR #141, opened 2018-11-06
feat(optOut): tag image with expiration_time:never

emjburns:
- Updating some of the formatting in `ResourceStateManager.updateState(event: Event)` so it's less nested.
- If the event is opt out, also tag the image with `"expiration_time":"never"` so that the opt out status is reflected on the image instead of just stored in the swabbie database.

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/Cacheable.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/Cacheable.kt
@@ -39,7 +39,7 @@ open class InMemoryCache<out T : Cacheable>(
 
   override fun contains(key: String?): Boolean {
     if (key == null) return false
-    return cache.get().find { it.name == key } != null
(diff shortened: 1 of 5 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/Cacheable.kt
```diff
@@ -39,7 +39,7 @@ open class InMemoryCache<out T : Cacheable>(
 
   override fun contains(key: String?): Boolean {
     if (key == null) return false
-    return cache.get().find { it.name == key } != null
+    return get().find { it.name == key } != null
```
robzienert: Are you sure?

emjburns: Pretty sure. This makes it use the `get()` function below, so that if the cache is null it will refresh the cache then return the cache. Found this out while testing.

PR #156, opened 2018-11-15
feat(grouping): add grouping (like app) to resources

emjburns:
We can't parse app name from image name. We can only parse package name.

This PR introduces a "grouping" field on an excludable (an image, an application), and defines the value and what type it is. For example, a deb called "spinredis-server-0.0.1" would be grouped by the `PACKAGE_NAME` `spinredis-server`. Because this grouping is not `APPLICATION`, the `Front50ApplicationExclusionPolicy` cannot provide any ownership about it, and now returns null (null == should not exclude).

This grouping is used when we group resources together (to send notifications) and to submit task to orca (for deletion, now all tasks will show up in the swabbie app).

For things that don't have grouping (resources marked before this PR is merged), tasks will appear under "swabbie" app and emails will be sent with no "application". This is fine, as we're sending all notifications to the "default owner" for now, until we can turn off notifications for emails entirely.

This PR also reformats some more kotlin-style nested blocks into more java-style intermediate variables.

### DIFF
--- a/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/securitygroups/AmazonSecurityGroupHandler.kt
+++ b/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/securitygroups/AmazonSecurityGroupHandler.kt
@@ -29,6 +29,7 @@ import com.netflix.spinnaker.swabbie.orca.OrcaJob
 import com.netflix.spinnaker.swabbie.orca.OrcaService
 import com.netflix.spinnaker.swabbie.orca.OrchestrationRequest
 import com.netflix.spinnaker.swabbie.repository.*
+import com.netflix.spinnaker.swabbie.utils.ApplicationUtils
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.stereotype.Component
 import java.time.Clock
@@ -52,7 +53,8 @@ class AmazonSecurityGroupHandler(
   private val securityGroupProvider: ResourceProvider<AmazonSecurityGroup>,
   private val orcaService: OrcaService,
   private val taskTrackingRepository: TaskTrackingRepository,
-  private val resourceUseTrackingRepository: ResourceUseTrackingRepository
+  private val resourceUseTrackingRepository: ResourceUseTrackingRepository,
+  private val applicationUtils: ApplicationUtils
 ) : AbstractResourceTypeHandler<AmazonSecurityGroup>(
   registry,
   clock,
@@ -78,7 +80,7 @@ class AmazonSecurityGroupHandler(
           log.info("This resource is about to be deleted {}", markedResource)
           orcaService.orchestrate(
             OrchestrationRequest(
-              application = determineApp(resource),
+              application = applicationUtils.determineApp(resource),
               job = listOf(
                 OrcaJob(
                   type = "deleteSecurityGroup",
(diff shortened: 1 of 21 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/agents/ScheduledAgent.kt
```diff
@@ -76,14 +77,25 @@ abstract class ScheduledAgent(
...
-         startupExecutorService.shutdown()
+         shutdown(startupExecutorService)
        }
      } catch (e: Exception) {
        log.error("Failed while waiting for cache to start in ${javaClass.simpleName}.", e)
      }
     }, 0, 5, TimeUnit.SECONDS)
   }
 
+  private fun shutdown(executorService: ExecutorService) {
```
emjburns: Introducing a more graceful shutdown.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/exclusions/ExclusionPolicy.kt
```diff
@@ -51,23 +51,26 @@ interface ExclusionPolicy {
...
     excludable: Excludable,
     exclusionType: ExclusionType = getType()
   ): String? {
-    keysAndValues(exclusions, exclusionType).let { kv ->
-      kv.values.toSet().flatten().let { exclusionValues ->
-        if (exclusionValues.size == 1 && exclusionValues[0] == "\\*") {
-          return wildcardMatchMessage(excludable.name!!, exclusionValues.toSet())
-        }
-      }
+    val kv: Map<String, List<String>> = keysAndValues(exclusions, exclusionType)
```
emjburns: Un-nesting the kotlin for readability.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
```diff
@@ -74,6 +78,20 @@ abstract class Resource : Excludable, Timestamped, HasDetails() {
   override fun toString(): String {
     return details.toString()
   }
+
+  fun getTagValue(key: String): String? {
```
emjburns: appversion (what we parse for the package name for images) is stored as an amazon tag on an image

swabbie-front50/src/main/kotlin/com/netflix/spinnaker/swabbie/front50/Front50ApplicationExclusionPolicy.kt
```diff
@@ -43,11 +52,14 @@ class Front50ApplicationExclusionPolicy(
   }
 
   override fun getType(): ExclusionType = ExclusionType.Application
+
   override fun apply(excludable: Excludable, exclusions: List<Exclusion>): String? {
-    keysAndValues(exclusions, ExclusionType.Allowlist).let { kv ->
-      findApplication(excludable, kv.values.flatten().toSet())?.let {
-        return byPropertyMatchingResult(exclusions, it)
-      }
+    val kv = keysAndValues(exclusions, ExclusionType.Allowlist)
```
emjburns: Un-nesting kotlin for readability

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/securitygroups/AmazonSecurityGroupHandler.kt
```diff
@@ -109,6 +109,14 @@ class AmazonSecurityGroupHandler(
     }
   }
 
+  private fun determineApp(resource: Resource): String {
```
asher: Can this be moved to the abstract class to avoid duplication across handlers?

emjburns: Good idea - I'll do that.

PR #160, opened 2018-11-20
chore(timeout) bump orca task timeout for delete

emjburns:
Right now, the orca tasks are timing out when we delete images. If the tasks fail, we never increment the metric that says we've deleted the images (because the `DeleteResourceEvent` is never fired: https://github.com/spinnaker/swabbie/blob/master/swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt#L141).

I want this metric to be published, and the resources to be removed from the marked list right after they're deleted. Should I:
- bump the timeout (this PR does that) so that we monitor the task for longer
- publish the event anyways even if the orca task fails (or parse more of the orca json to figure out if it was a timeout failure, and only publish the event on timeout failures)
- not care about tracking the orca task at all, because if the resource didn't get deleted because of a transient orca failure we would just try again next time? 

(also simplifying a log message in this PR)

### DIFF
--- a/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
+++ b/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
@@ -91,7 +91,7 @@ class AmazonImageHandler(
               "imageIds" to markedResources.map { it.resourceId }.toSet(),
               "cloudProvider" to AWS,
               "region" to workConfiguration.location,
-              "stageTimeoutMs" to Duration.ofMinutes(15).toMillis().toString()
+              "stageTimeoutMs" to Duration.ofMinutes(40).toMillis().toString()
             )
           )
         ),
--- a/swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt
+++ b/swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt
@@ -143,10 +143,11 @@ class OrcaTaskMonitoringAgent (
             }
             response.status.isFailure() -> {
               log.error(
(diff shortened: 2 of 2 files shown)


swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
```diff
@@ -91,7 +91,7 @@ class AmazonImageHandler(
               "imageIds" to markedResources.map { it.resourceId }.toSet(),
               "cloudProvider" to AWS,
               "region" to workConfiguration.location,
-              "stageTimeoutMs" to Duration.ofMinutes(15).toMillis().toString()
+              "stageTimeoutMs" to Duration.ofMinutes(40).toMillis().toString()
```
asher: This is fine and correct, in that Orca's `MonitorDeleteImageTask` is a `OverridableTimeoutRetryableTask` which will honor the longer stageTimeoutMs parameter. However, the swabbie -> orca -> clouddriver trifecta may pose major operational challenges in an environment where there are > 100k images to delete, orca and clouddriver are resource constrained, AWS describeImage calls are extremely slow, and swabbie activity is limited to what happens to be peak spinnaker usage hours.

* `MonitorDeleteImageTask` has a backoff period of 5 seconds. If these may actually run for ~30 minutes, we can calculate how many can be queued concurrently based on current capacity of orca and our clouddriver shard for orchestrations.
* Consider making image deletion fire-and-forget from the perspective of Swabbie. This could preserve the existing orca stage (minus the monitor task) and clouddriver deleteImage kato task. Swabbie could optionally mark a resource as "deleting" and reconcile as deleted or re-send a delete request on a future crawl of all resources. I prefer this approach over the current PR.
* If there are major blockers to a fire-and-forget approach, we may need to consider running a dedicated orca cluster/queue for swabbie, and figure out clouddriver resourcing for it.

PR #165, opened 2018-11-27
feat(fastProps): work and maxBatchSize fast props

emjburns:
Have swabbie work decision backed by a fast property (`swabbie.work` as boolean).
Have the max batch size backed by fast properties (`${namespace}.maxItemsProcessedPerCycle`)


cfieber:
LGTM

One thing to maybe consider is pushing the DynamicConfigService inside the Swabbie ConfigProperties bean.

I did something similar in orca and felt like it worked out okay:
https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/locks/LockingConfigurationProperties.java#L10

The nice side effect being that you only have one bean to inject all the places where you are making configuration driven decisions.

Totally optional though, just a thought!

emjburns:
@cfieber I'll think about that, and if I want to have more places where we use dynamic config service I might refactor. Thanks for linking the class as reference!

PR #167, opened 2018-11-28
feat(deleteTime): recalculate time for some resources

emjburns:
We may want to update the deletion time of a bunch of resources, like if we want to speed up or slow down deletion of resources.

This admin endpoint lets you update the deletion time for the oldest `numResources` resources to a value of `retentionS` seconds from now. It operates in a single namespace.

### DIFF
--- a/swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/ResourceTypeHandlerTest.kt
+++ b/swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/ResourceTypeHandlerTest.kt
@@ -48,6 +48,7 @@ import java.time.Clock
 import java.time.Instant
 import java.time.ZoneOffset
 import java.util.*
+import java.util.concurrent.TimeUnit
 
 object ResourceTypeHandlerTest {
   private val resourceRepository = mock<ResourceTrackingRepository>()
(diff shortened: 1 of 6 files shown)


swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/ResourceController.kt
```diff
@@ -196,4 +196,22 @@ class ResourceController(
...
+
+  /**
+   * Recalculate deletion timestamp to @retentionS seconds in the future for the
+   * oldest @numResources that are marked
+   */
+  @RequestMapping(value = ["/recalculate/{namespace}/"], method = [RequestMethod.PUT])
+  fun recalculate(
+      @PathVariable namespace: String,
+      @RequestParam(required = true) retentionS: Long,
+      @RequestParam(required = false, defaultValue = "100") numResources: Int
```
ajordens: Is it safer to just not have a default value for `numResources`? Doesn't seem like a stretch to ask operators to be explicit about what they're like done?

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/ResourceTypeHandlerTest.kt
```diff
@@ -659,6 +660,117 @@ object ResourceTypeHandlerTest {
     defaultHandler.deletionTimestamp(configuration) shouldMatch equalTo(plus2Days)
   }
 
+  @Test
+  fun `update delete timestamp should work`() {
+    val timestamp10days = clock.millis().plus(TimeUnit.DAYS.toMillis(10))
+    val timestamp1hour = clock.millis().plus(TimeUnit.HOURS.toMillis(1))
+    val resource1 = TestResource(
```
ajordens: test looks good functionally ... there might be some opportunity to reduce the verbosity of the setup.

```
val resource1 = appResource("1", "testResource-v001", app("testResource"), DAYS.toMillis(11))
val resource4 = imageResource("4", "my-package-0.0.2", package("my-package"), DAYS.toMillis(11))
```

motivation being to cut down on the amount of repeated boiler plate thus making it easier to grok the actual test intent.

swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/ResourceController.kt
```diff
@@ -196,4 +196,22 @@ class ResourceController(
         && workConfiguration.location == namespace.region
     } ?: throw NotFoundException("No configuration found for $namespace")
   }
+
+  /**
+   * Recalculate deletion timestamp to @retentionS seconds in the future for the
```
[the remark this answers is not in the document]
emjburns: It would seem so (https://kotlinlang.org/docs/reference/kotlin-doc.html#inline-markup).

Updating both!

swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/ResourceController.kt
emjburns: Sure, that makes sense! Changing.

swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/ResourceController.kt
```diff
@@ -196,4 +196,22 @@ class ResourceController(
...
+   * oldest @numResources that are marked
+   */
+  @RequestMapping(value = ["/recalculate/{namespace}/"], method = [RequestMethod.PUT])
```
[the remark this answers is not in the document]
emjburns: Created a new /admin controller with this endpoint to make it explicit. Also pulled some of the common functionality (like getting a workConfiguration or handler) into a ControllerUtils class.

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/ResourceTypeHandlerTest.kt
emjburns: Going to refactor this whole test class in a separate PR. I agree it's hard to read, and there is way too much boilerplate.

ajordens: 👍

PR #168, opened 2018-11-29
feat(agents): configurable max num of concurrent agents

emjburns:
Right now swabbie instances are greedy and will grab a ton of locks at once. That's not good when we have 4 regions to delete from, with 12 different agents (a mark, notify, and delete agent for each region).

This PR introduces a configurable max number of locks. This really only applies to the mark/notify/delete agents - you'll notice that the orcaTaskMonitoringAgent doesn't respect this max. That's ok, because it doesn't do much work.

I'm not sure if this is the right way to go about limiting. This seemed easier than modifying the lockManager in Kork to have a max locks per owner...

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/LockingService.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/LockingService.kt
@@ -18,31 +18,12 @@ package com.netflix.spinnaker.swabbie
 
 import com.netflix.spinnaker.config.LockingConfigurationProperties
 import com.netflix.spinnaker.kork.lock.LockManager
-import org.slf4j.Logger
-import org.slf4j.LoggerFactory
 import java.time.Duration
-import java.util.concurrent.atomic.AtomicInteger
 
 class LockingService(
   private val lockManager: LockManager,
   private val lockingConfigurationProperties: LockingConfigurationProperties
 ) : LockManager by lockManager {
-  private val log: Logger = LoggerFactory.getLogger(javaClass)
-
-  val numLocksCurrentlyHeld = AtomicInteger(0)
-
   val swabbieMaxLockDuration: Duration
     get() = Duration.ofMillis(lockingConfigurationProperties.maximumLockDurationMillis!!)
-
-  val swabbieMaxConcurrentLocks: Int
-    get() = lockingConfigurationProperties.maxConcurrentLocks
-
-  @Synchronized override fun acquireLock(lockOptions: LockManager.LockOptions, onLockAcquiredCallback: Runnable): LockManager.AcquireLockResponse<Void> {
-    if (numLocksCurrentlyHeld.get() >= lockingConfigurationProperties.maxConcurrentLocks) {
-      log.warn("Too many locks are held " +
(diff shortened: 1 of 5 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/LockingService.kt
```diff
@@ -18,12 +18,31 @@ package com.netflix.spinnaker.swabbie
...
+
+  val swabbieMaxConcurrentLocks: Int
+    get() = lockingConfigurationProperties.maxConcurrentLocks
+
+  @Synchronized override fun acquireLock(lockOptions: LockManager.LockOptions, onLockAcquiredCallback: Runnable): LockManager.AcquireLockResponse<Void> {
+    if (numLocksCurrentlyHeld.get() >= lockingConfigurationProperties.maxConcurrentLocks) {
+      log.warn("Too many locks are held " +
+          "(held=${numLocksCurrentlyHeld.get()}, wanted=${lockingConfigurationProperties.maxConcurrentLocks}).")
+    }
+    numLocksCurrentlyHeld.incrementAndGet()
```
cfieber: probably want to only increment this if the acquireLock is successful

cfieber: is there nothing explicitly calling releaseLock on these locks? (I could see that, assuming the model is acquire the lock then let it TTL away before the next instance can acquire it)

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/LockingService.kt
```diff
@@ -18,12 +18,31 @@ package com.netflix.spinnaker.swabbie
...
+
+  val numLocksCurrentlyHeld = AtomicInteger(0)
+
   val swabbieMaxLockDuration: Duration
     get() = Duration.ofMillis(lockingConfigurationProperties.maximumLockDurationMillis!!)
+
+  val swabbieMaxConcurrentLocks: Int
+    get() = lockingConfigurationProperties.maxConcurrentLocks
+
+  @Synchronized override fun acquireLock(lockOptions: LockManager.LockOptions, onLockAcquiredCallback: Runnable): LockManager.AcquireLockResponse<Void> {
```
cfieber: I wonder if you could take advantage of the fact that there is an onLockAcquiredCallback here and wrap something around that to do a try { increment } finally { decrement }

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/LockingService.kt
emjburns: No, from my sleuthing it seems they're being called within the redisLockManager. I might be able to override that also, but was having trouble. Because decrementing there would be better than in the `finally` of each function.

emjburns:
@asher thanks for the much much better solution. Made the number of threads the agent executor has  tunable. Defaulted it to 2 for right now. This seems to fix the problem much more elegantly. Will revisit this default number/approach as I test.

emjburns:
Also setting the orca monitoring agent to run less often, since it's so chatty

PR #174, opened 2019-01-04
feat(notify): notifications not always required

emjburns:
Some resources need to be cleaned too often to notify users every time we delete something. Janitor monkey did not notify for image clean up, so this PR adds that functionality into swabbie.

The default behavior is the same. However, if you set 
``` 
notifications:
  enabled: true
  required: false
```
the notification agent will run but no notifications will be sent. 

This behavior is a little bit convoluted - a cleaner solution would be to have deletion not depend on notifying for all resources. However, that would be a large change to the assumptions of Swabbie (something must be notified before it was deleted. This PR keeps the same assumptions and adds support for not notifying with minimal changes.

There are also some indentation changes because I ran the auto-formatter.

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
@@ -61,8 +61,25 @@ class CloudProviderConfiguration {
   }
 }
 
+/**
+ * Notifications have both an [enabled] and a [required] configuration.
+ *  [enabled] is whether or not the notification agent runs for this configuration.
+ *  [required] is whether or not a resource must be notified before it can be deleted.
+ *
+ * If [enabled] is true and [required] is true then notifications will be sent to the owner.
+ *
+ * If [enabled] is true and [required] is false a resource will be marked as 'notification not needed'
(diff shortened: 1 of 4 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
```diff
@@ -61,8 +61,22 @@ class CloudProviderConfiguration {
...
+ *  [enabled] is whether or not the notification agent runs for this configuration.
+ *  [required] is whether or not a resource must be notified before it can be deleted.
+ *
+ * If [enabled] is true and [required] is true then notifications will be sent to the owner.
+ *
+ * If [enabled] is true and [required] is false a resource will be marked as 'notification not needed'
+ *  and will be able to be deleted. No notifications will be sent.
+ *
+ * If [enabled] is false then the notification agent will not run for this configuration
+ *  and resources of this type won't be able to be deleted.
```
asher: I'd leave a TODO here for removing the hard dependency between resource and notification agents but think this PR makes sense as an interim way forward.

emjburns: Ok will add. Thanks!

PR #179, opened 2019-03-02
feat(snapshots): snapshot handler to first pass

emjburns:
Mark snapshots. This will only mark things that fit the pattern of auto-snapshotted during a netflix bake. 

There's also a change to the image handler. Some background: AMIs contain a mapping to snapshots. Snapshots do not contain the mapping of what image they're a snapshot of. The change: as we go through the AMIs we record what snapshots they have, with at TTL of 2 days. We use this information in the snapshot handler to not delete snapshots which have been recorded here aka the image still exists for this snapshot.

Coming: deletion of snapshots via an orca task. It's marked as to-do right now.

An important note: you can't delete a snapshot of an ebs volume when the image it was taken from exists. So, there is really no harm in 'accidentally' deleting the snapshot.

### DIFF
--- a/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/edda/providers/EddaAmazonSnapshotProvider.kt
+++ b/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/edda/providers/EddaAmazonSnapshotProvider.kt
@@ -1,4 +1,5 @@
 /*
+ *
  * Copyright 2018 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License")
@@ -12,6 +13,7 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
  */
 
 package com.netflix.spinnaker.swabbie.aws.edda.providers
@@ -21,63 +23,63 @@ import com.netflix.spinnaker.config.EddaApiClient
 import com.netflix.spinnaker.kork.core.RetrySupport
 import com.netflix.spinnaker.swabbie.Parameters
 import com.netflix.spinnaker.swabbie.ResourceProvider
-import com.netflix.spinnaker.swabbie.aws.images.AmazonImage
 import com.netflix.spinnaker.swabbie.aws.edda.EddaService
-import com.netflix.spinnaker.swabbie.model.IMAGE
+import com.netflix.spinnaker.swabbie.aws.snapshots.AmazonSnapshot
+import com.netflix.spinnaker.swabbie.model.SNAPSHOT
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
 import retrofit.RetrofitError
 
 @Component
-open class EddaAmazonImageProvider(
+open class EddaAmazonSnapshotProvider(
   eddaApiClients: List<EddaApiClient>,
   private val retrySupport: RetrySupport,
   private val registry: Registry
-) : ResourceProvider<AmazonImage>, EddaApiSupport(eddaApiClients, registry) {
+) : ResourceProvider<AmazonSnapshot>, EddaApiSupport(eddaApiClients, registry) {
   private val log: Logger = LoggerFactory.getLogger(javaClass)
 
-  override fun getAll(params: Parameters): List<AmazonImage>? {
+  override fun getAll(params: Parameters): List<AmazonSnapshot>? {
     withEddaClient(
       region = params.region,
       accountId = params.account,
       environment = params.environment
     )?.run {
-      return getAmis()
+      return getSnapshots()
     }
 
     return emptyList()
   }
 
-  override fun getOne(params: Parameters): AmazonImage? {
+  override fun getOne(params: Parameters): AmazonSnapshot? {
     withEddaClient(
       region = params.region,
       accountId = params.account,
       environment = params.environment
     )?.run {
-      return getAmi(params.id)
+      return getSnapshot(params.id)
     }
 
     return null
   }
 
-  private fun EddaService.getAmis(): List<AmazonImage> {
+  private fun EddaService.getSnapshots(): List<AmazonSnapshot> {
     return try {
       retrySupport.retry({
-        this.getImages()
+        this.getSnapshots()
       }, maxRetries, retryBackOffMillis, true)
     } catch (e: Exception) {
-      registry.counter(eddaFailureCountId.withTags("resourceType", IMAGE)).increment()
-      log.error("failed to get images", e)
+      registry.counter(eddaFailureCountId.withTags("resourceType", SNAPSHOT)).increment()
+      log.error("failed to get snapshots", e)
       throw e
     }
   }
 
-  private fun EddaService.getAmi(imageId: String): AmazonImage? {
+  private fun EddaService.getSnapshots(snapshotId: String): AmazonSnapshot? {
     return try {
       retrySupport.retry({
         try {
-          this.getImage(imageId)
+          this.getSnapshot(snapshotId)
         } catch (e: Exception) {
           if (e is RetrofitError && e.response.status == 404) {
             null
@@ -87,8 +89,8 @@ open class EddaAmazonImageProvider(
         }
       }, maxRetries, retryBackOffMillis, false)
     } catch (e: Exception) {
-      registry.counter(eddaFailureCountId.withTags("resourceType", IMAGE)).increment()
-      log.error("failed to get image {}", imageId, e)
+      registry.counter(eddaFailureCountId.withTags("resourceType", SNAPSHOT)).increment()
+      log.error("failed to get snapshot {}", snapshotId, e)
       throw e
     }
   }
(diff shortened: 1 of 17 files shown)


swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImage.kt
```diff
@@ -27,9 +28,24 @@ data class AmazonImage(
...
   val state: String,
+  val blockDeviceMappings: List<AmazonBlockDevice>?,
   override val resourceId: String = imageId,
   override val resourceType: String = IMAGE,
   override val cloudProvider: String = AWS,
   override val name: String?,
   private val creationDate: String?
 ) : AmazonResource(creationDate)
+
+@JsonTypeName("blockDevice")
```
jeyrschabu: This annotation is not necessary here unless we are intending to operate on objects of this kind in the future.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImage.kt
```diff
@@ -27,9 +28,24 @@ data class AmazonImage(
...
+data class AmazonBlockDevice(
+  val ebs: EbsBlockDevice?,
+  val noDevice: String?,
+  val deviceName: String?,
+  val virtualName: String?
+)
+
+@JsonTypeName("ebs")
```
jeyrschabu: Same comment as `blockDevice`.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/edda/providers/EddaAmazonSnapshotProvider.kt
```diff
@@ -0,0 +1,97 @@
+/*
+ *
+ * Copyright 2018 Netflix, Inc.
```
jeyrschabu: 2019. Time flies

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImage.kt
jeyrschabu: Are we seeing failures to marshal an image when `JsonTypeName` is omitted?

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
```diff
@@ -155,9 +156,17 @@ class AmazonImageHandler(
...
-        it.set(NAIVE_EXCLUSION, true) // exclude these with exclusions
+    images.forEach { image ->
+      if (image.name == null || image.description == null) {
+        image.set(NAIVE_EXCLUSION, true) // exclude these with exclusions
+      }
+
+      if (image.blockDeviceMappings != null){
+        image.blockDeviceMappings.forEach { blockDevice ->
+          if (blockDevice.ebs != null && blockDevice.ebs.snapshotId != null) {
+            usedSnapshotRepository.recordUse(blockDevice.ebs.snapshotId, "aws:${params.region}:${params.account}") //todo eb: is this value needed?
```
jeyrschabu: For reference, does it make sense to record which image is linked to a snapshot? I personally don't see a huge value in doing so, but could be used if we ever want to create a correlation matrix for different resource types. Just a thought, not something required here.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/snapshots/AmazonSnapshotHandler.kt
```diff
@@ -0,0 +1,151 @@
...
+        environment = workConfiguration.account.environment
+      ))
+    return candidates
+  }
+
+  private fun checkIfImagesExist(candidates: List<AmazonSnapshot>, params: Parameters) {
+    log.info("Checking for existing images for {} snapshots. Params: {} ", candidates.size, params)
+    candidates.forEach { snapshot ->
+      if (snapshot.description == null) {
+        snapshot.set(SNAPSHOT_NOT_FROM_BAKE, true)
```
jeyrschabu: Alternatively, we can use the generic `NAIVE_EXCLUSION` classifier here. This avoids tying to a netflix only pattern.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/Dates.kt
```diff
@@ -50,6 +52,12 @@ class Dates {
 
       throw exception!!
     }
+
+    fun toCreationDate(timestampMillis: Long) : String {
```
jeyrschabu: Since this function is not specific to creation dates, it makes sense to rename it to something more generic for dates.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImage.kt
emjburns: Good point - I was seeing some failures and was trying to figure out why. I think the root of my problem was that I was using an object mapper in the tests which didn't have the json ignore unknown properties set. I will test out removing these annotations.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
emjburns: I think it's useful (usecase: when we delete an image we could do some sort of 'restore' with the snapshot eventually). 

That being said, I'd prefer to refactor later and build something with that specific purpose in mind. This is just meant to build a mapping because aws only stores the mapping of image -> snapshot, and we need to know whether the image is in use.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/snapshots/AmazonSnapshotHandler.kt
emjburns: Can you explain more? What general set of things does `NAIVE_EXCLUSION` capture?

emjburns:
@jeyrschabu any suggestions on how to make this memoization more generic?

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
jeyrschabu: Discussed offline. To summarized, there will be a PR addressing a general approach to cache references, avoiding the need for a `Foo` resource to need a `FooInUseRepository`. Ideally we would keep an`InUseRepository` to handle in use resource references for resources.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/snapshots/AmazonSnapshotHandler.kt
jeyrschabu: `NAIVE_EXCLUSION` is a catch-all exclusion classifier that would make the resource excludable by default.

emjburns:
@jeyrschabu and I discussed offline and I'm going to refactor to make a general 'used' cache. This is an easy change to generalize the case where one resource knows if another resources is used. If more specific features are necessary those can be done in a later PR.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/edda/providers/EddaAmazonSnapshotProvider.kt
emjburns: I've fixed my header for the future LOL

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/Dates.kt
emjburns: It takes a timestamp and changes it to the 'aws creation date' format. I think the name is fine, given that that is the functionality that it is doing.

swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/snapshots/AmazonSnapshotHandlerTest.kt
```diff
@@ -287,6 +287,22 @@ object AmazonSnapshotHandlerTest {
     verify(resourceRepository, times(0)).upsert(any(), any())
   }
 
+  @Test
+  fun `should not mark snapshots if they're in use`() {
```
jeyrschabu: Nice!

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisUsedResourceRepository.kt
```diff
@@ -0,0 +1,58 @@
...
+
+@Component
+class RedisUsedResourceRepository(
+  redisClientSelector: RedisClientSelector
+) : UsedResourceRepository {
+
+  private val redisClientDelegate: RedisClientDelegate = redisClientSelector.primary("default")
+  private val log = LoggerFactory.getLogger(javaClass)
+
+  private val expTime = TimeUnit.DAYS.toSeconds(2).toInt()
```
asher: This should probably be configurable.

emjburns: Done

PR #181, opened 2019-04-05
Pilot use of Kork BOM for dependencies

robfletcher:


build.gradle
```diff
@@ -20,46 +20,20 @@ buildscript {
...
-  if (localVersions) {
-    logger.info("Enabling mavenLocal repo for $localVersions")
-    repositories {
-      mavenLocal()
-    }
-  }
-
-  spinnaker {
-    dependenciesVersion = spinnakerDependenciesVersion
-  }
```
robfletcher: No longer need any of this spinnaker dependency version stuff

build.gradle
```diff
@@ -77,6 +52,9 @@ subprojects {
     jcenter()
   }
 
+  dependencies {
+    api(platform("com.netflix.spinnaker.kork:kork-bom:4.0.0-rc.8-bom"))
+  }
```
robfletcher: apply the BOM to all sub-projects

gradle/kotlin.gradle
```diff
@@ -18,16 +18,6 @@ apply plugin: "nebula.kotlin"
...
-    jvmTarget = "1.8"
-  }
-}
-configurations.all {
-  resolutionStrategy {
-    eachDependency { details ->
-      if (details.requested.group == "org.jetbrains.kotlin") {
-        details.useVersion kotlinVersion
-      }
-    }
```
robfletcher: Kotlin's bom (which we get transitively from `kork-bom`) does this alignment now.

swabbie-redis/src/test/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceUseTrackingRepositoryTest.kt
```diff
@@ -27,7 +27,7 @@ import com.netflix.spinnaker.kork.jedis.EmbeddedRedis
 import com.netflix.spinnaker.kork.jedis.JedisClientDelegate
 import com.netflix.spinnaker.kork.jedis.RedisClientSelector
 import com.netflix.spinnaker.swabbie.test.TestResource
-import main.java.com.netflix.spinnaker.kork.test.time.MutableClock
+import com.netflix.spinnaker.kork.test.time.MutableClock
```
robfletcher: This seemed to just be wrong. Possibly because I'm pulling in a newer Kork. ¯\_(ツ)_/¯

build.gradle
robzienert: omg

build.gradle
```diff
@@ -20,46 +20,20 @@ buildscript {
...
-      checkLocalVersions.putAll(extVers)
-    }
-  }
-
-  def localVersions = checkLocalVersions.findAll { it.value.endsWith('-SNAPSHOT') }
-  if (localVersions) {
-    logger.info("Enabling mavenLocal repo for $localVersions")
-    repositories {
-      mavenLocal()
-    }
```
robzienert: wont we need something like this if kork is published locally?

robfletcher: Yeah, that's true. I'll give it some thought. Right now I can't actually publish kork locally because of fucking Javadoc so I'll probably try to fix that first.

robfletcher: Basically you'd want to enable `mavenLocal()` if the `kork-bom` version ends in `-SNAPSHOT`. Should be pretty easy.

PR #185, opened 2019-04-26
Add metrics for number of candidates processed

aravindmd:
- We log the summary of number of candidates processed , adding these numbers as metrics to get better visibility into the health of the system

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
@@ -442,7 +442,10 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
     action: Action
   ) {
     val totalProcessed = Math.min(getMaxItemsProcessedPerCycle(workConfiguration), totalResourcesVisitedCounter.get())
-    registry.counter(numberOfCandidates.withTags("actionName", action.name)).increment()
+    registry.counter(numberOfCandidates.withTags(
+      "actionName", action.name,
+      "candidates", candidateCounter.get().toString()
+    )).increment()
     log.info("${action.name} Summary: {} candidates out of {} processed. {} scanned. {} excluded. Configuration: {}",
       candidateCounter.get(),
       totalProcessed,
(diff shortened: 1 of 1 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -440,6 +442,10 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
     action: Action
   ) {
     val totalProcessed = Math.min(getMaxItemsProcessedPerCycle(workConfiguration), totalResourcesVisitedCounter.get())
+    registry.counter(numberOfCandidates.withTags(
```
emjburns: We should also add type of resource in here

aravindmd: Makes sense , added the resource type

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -440,6 +442,11 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
...
+      "actionName", action.name,
+      "resourceType", workConfiguration.resourceType,
+      "candidates", candidateCounter.get().toString()
+    )).increment()
```
emjburns: I don't think this is the correct way to do this. This will publish a metric with a tag of the value, which is not what we want. I think we want to use a [PolledMeter](https://github.com/Netflix/spectator/blob/master/spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledMeter.java#L79)

emjburns: (or more realistically a gauge)

jeyrschabu: +1 a gauge is more appropriate than a simple counter here.

PR #188, opened 2019-05-02
fix(core): Ensuring maxItemsToProcess is respected

jeyrschabu:
- fix candidateCounter with respect to maxItemsToProcess

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
@@ -280,7 +280,7 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
       val maxItemsToProcess = Math.min(preProcessedCandidates.size, getMaxItemsProcessedPerCycle(workConfiguration))
 
       for (candidate in preProcessedCandidates) {
-        if (candidateCounter.get() > maxItemsToProcess) {
+        if (candidateCounter.get() >= maxItemsToProcess) {
           log.info("Max items ({}) to process reached, short-circuiting", maxItemsToProcess)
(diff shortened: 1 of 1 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -280,7 +280,7 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
       val maxItemsToProcess = Math.min(preProcessedCandidates.size, getMaxItemsProcessedPerCycle(workConfiguration))
 
       for (candidate in preProcessedCandidates) {
-        if (candidateCounter.get() > maxItemsToProcess) {
+        if (candidateCounter.get() >= maxItemsToProcess) {
```
jeyrschabu: Makes sure to exit the loop when the maxItemsToProcess is reached.

jeyrschabu:
Future state, I am thinking about handling counting inside of a `CycleStatCounter` or something similar (naming is hard). This would do for now.

jeyrschabu:
Fixing tests

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
emjburns: HA 😄

emjburns:
@jeyrschabu what's your thinking behind handling the counting somewhere else?

jeyrschabu:
@emjburns right now we have a few counters tracking candidates (mark|delete|excluded|notified). They are a bit scattered and easy to miss. I am thinking along the lines of consolidating these counters and make them easy to assert in test.

PR #189, opened 2019-05-03
fix(*): Override hashcode, equals in resource data class

jeyrschabu:
- add explicit equals and hashcode to data classes
- fix a bug on Resource.equals
- use TestResource from test where possible instead of creating new ones

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
@@ -69,7 +69,7 @@ abstract class Resource : Excludable, Timestamped, HasDetails() {
       return true
     }
     return other is Resource &&
-      other.resourceId == this.resourceId && other.resourceType == this.resourceId
+      other.resourceId == this.resourceId && other.resourceType == this.resourceType
   }
 
   override fun hashCode(): Int {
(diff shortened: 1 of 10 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
```diff
@@ -69,7 +69,7 @@ abstract class Resource : Excludable, Timestamped, HasDetails() {
       return true
     }
     return other is Resource &&
-      other.resourceId == this.resourceId && other.resourceType == this.resourceId
+      other.resourceId == this.resourceId && other.resourceType == this.resourceType
```
jeyrschabu: lol

jeyrschabu: This is what got me thinking about the equals/hashcode. Obviously a bug otherwise all equality checks would fail.
Instead the default `equals` on the data class would be called. Explains why we haven't run into trouble.

swabbie-front50/src/test/kotlin/com/netflix/spinnaker/swabbie/front50/ApplicationResourceOwnerResolutionStrategyTest.kt
```diff
@@ -42,16 +41,12 @@ object ApplicationResourceOwnerResolutionStrategyTest {
         Application(name = "test", email = "test@netflix.com")
       )
 
-    subject.resolve(ResourceOne()) shouldMatch equalTo("name@netflix.com")
+    subject.resolve(
+      TestResource(
```
jeyrschabu: Using the `TestResource` class where applicable. Took the opportunity to do a minor cleanup here.

jeyrschabu:
Another tick was the fact that I notice some intermittent test failures. This PR should take care of that.

jeyrschabu:
Once this merged, I can merge this: https://github.com/spinnaker/swabbie/pull/188

PR #203, opened 2019-05-21
fix(header): Wrap calls to allow anonymous calls

aravindmd:
- Wrap calls to allow anonymous calls


jeyrschabu:
@aravindmd LGTM but we also should probably apply this to other places as well.

aravindmd:
I think i applied to most of the calls , i'll take another quick pass.

jeyrschabu:
@aravindmd take a look at all the `Edda*Provider` in the aws module

PR #210, opened 2019-06-12
chore(logs): changing level of some logs

emjburns:
I want to disable debug logs in production, so swept through the logs to and demoted some to debug.

Please let me know if you disagree with any of these.

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
@@ -168,7 +168,7 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
   override fun optOut(resourceId: String, workConfiguration: WorkConfiguration) {
     val resource = getCandidate(resourceId, "", workConfiguration)
       ?: return
-    log.debug("Opting out resource $resourceId in namespace ${workConfiguration.namespace}")
+    log.info("Opting out resource $resourceId in namespace ${workConfiguration.namespace}")
     val resourceWithOwner = listOf(resource).withResolvedOwners(workConfiguration).first()
     val newMarkedResource = MarkedResource(
       resource = resourceWithOwner,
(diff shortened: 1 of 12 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -437,7 +437,7 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
   ) {
     if (markedResource != null && !workConfiguration.dryRun) {
       try {
-        log.info("{} is no longer a candidate because: $reason.", markedResource.uniqueId())
+        log.debug("{} is no longer a candidate because: $reason.", markedResource.uniqueId())
```
aravindmd: This would be useful , right ?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/ResourceOwnerResolver.kt
```diff
@@ -70,7 +70,7 @@ open class ResourceOwnerResolver<in T : Resource>(
         null
       }
     } catch (e: Exception) {
-      log.info("Failed to find owner for {}.", resource, e)
+      log.warn("Failed to find owner for {}.", resource, e)
```
aravindmd: same here ?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
emjburns: this message hasn't been actually that useful in practice, I think.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/ResourceOwnerResolver.kt
emjburns: Warn will show.

PR #213, opened 2019-06-19
fix(schedule): Fix for swabbie scheduling

jeyrschabu:
- remove double scheduling after caches are loaded
- refactored ScheduledAgent#timeToWork to not include off hours
- if off hours are needed, the config can be explicitly updated

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
@@ -26,9 +26,9 @@ import java.time.LocalTime
 import java.time.ZoneId
 
 /**
- * @param minImagesUsedByLC minimum number of images used by launch configurations that should be
+ * @param {@link #minImagesUsedByLC} minimum number of images used by launch configurations that should be
  *  in the cache. If not present agents will error to prevent mass marking in the case of edda problems.
- * @param minImagesUsedByInst minimum number of images used by instances that should be
+ * @param {@link #minImagesUsedByInst} minimum number of images used by instances that should be
  *  in the cache. If not present agents will error to prevent mass marking in the case of edda problems.
  */
 @ConfigurationProperties("swabbie")
@@ -190,31 +190,16 @@ class Schedule {
   var enabled: Boolean = true
   var startTime: String = "09:00"
   var endTime: String = "16:00"
-  val allowedDaysOfWeek: MutableList<DayOfWeek> = mutableListOf()
-  var defaultTimeZone: String = "America/Los_Angeles"
 
-  fun getResolvedDays(): List<DayOfWeek> {
-    return if (!enabled) {
-      mutableListOf(
-        DayOfWeek.MONDAY,
-        DayOfWeek.TUESDAY,
-        DayOfWeek.WEDNESDAY,
-        DayOfWeek.THURSDAY,
-        DayOfWeek.FRIDAY,
-        DayOfWeek.SATURDAY,
-        DayOfWeek.SUNDAY
-      )
-    } else if (allowedDaysOfWeek.isEmpty()) {
-      mutableListOf(
-        DayOfWeek.MONDAY,
-        DayOfWeek.TUESDAY,
-        DayOfWeek.WEDNESDAY,
-        DayOfWeek.THURSDAY
-      )
-    } else {
-      allowedDaysOfWeek
-    }
-  }
+  // default days overridable in config
+  var allowedDaysOfWeek: List<DayOfWeek> = listOf(
+    DayOfWeek.MONDAY,
+    DayOfWeek.TUESDAY,
+    DayOfWeek.WEDNESDAY,
+    DayOfWeek.THURSDAY
+  )
+
+  var defaultTimeZone: String = "America/Los_Angeles"
 
   fun getResolvedStartTime(): LocalTime {
     return LocalTime.parse(startTime)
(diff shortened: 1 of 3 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
```diff
@@ -25,12 +25,6 @@ import java.time.LocalDateTime
 import java.time.LocalTime
 import java.time.ZoneId
 
-/**
```
emjburns: Why did you delete this comment? I think it's helpful.

jeyrschabu: I was feeling like deleting some comments I guess. Will put them back.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/agents/ScheduledAgent.kt
```diff
@@ -75,59 +76,49 @@ abstract class ScheduledAgent(
...
-      try {
-        while (!cacheStatus.cachesLoaded()) {
-          log.info("Can't work until the caches load...")
-          Thread.sleep(Duration.ofSeconds(5).toMillis())
-        }
-        log.info("Caches loaded.")
-        scheduleSwabbie()
-      } catch (e: Exception) {
-        log.error("Failed while waiting for cache to start in ${javaClass.simpleName}.", e)
+    try {
```
emjburns: Great catch :)

swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
emjburns: Thank you!

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/agents/ScheduledAgent.kt
```diff
@@ -75,59 +76,49 @@ abstract class ScheduledAgent(
...
-                timeToWork(swabbieProperties.schedule, clock) -> {
-                  if (dynamicConfigService.isEnabled("swabbie.work", true)) {
-                    log.info("Swabbie schedule: working")
-                    runSwabbie()
-                  } else {
-                    log.info("Swabbie schedule: disabled via property swabbie.work")
-                  }
+    runningTask = executorService.scheduleWithFixedDelay({
+      when {
+        dynamicConfigService.isEnabled(SWABBIE_FLAG_PROPERY, true) -> {
```
emjburns: This condition seems to be `if swabbie.work == true then do not work`. Based on the log message, this seems to be flipped

jeyrschabu: Yeah good catch. I caught that in the other PR I was working last night. Will update

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/agents/ScheduledAgent.kt
```diff
@@ -168,21 +159,14 @@ abstract class ScheduledAgent(
...
-      val now = LocalTime.from(clock.instant().atZone(schedule.getZoneId()))
-
-      return if (startTime <= endTime) {
-        // Work during the day
-        now.isAfter(startTime) && now.isBefore(endTime)
-      } else {
-        // Work during the night
-        now.isAfter(startTime) || now.isBefore(endTime)
-      }
+      return if (schedule.enabled && LocalDate.now(clock).dayOfWeek in schedule.allowedDaysOfWeek) {
```
emjburns: I personally don't like lifting the return out of the if statement because I feel like it pulls us closer to the groovy "oh the last line is implicitly the return of the function, and I find that hard to easily parse when reading code. I prefer to keep the two returns.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/agents/ScheduledAgent.kt
```diff
@@ -168,21 +159,14 @@ abstract class ScheduledAgent(
...
+        val startTime: LocalTime = schedule.getResolvedStartTime()
+        val endTime: LocalTime = schedule.getResolvedEndTime()
+        val now = LocalTime.from(clock.instant().atZone(schedule.getZoneId()))
+
+        startTime.isBefore(endTime) &&
```
emjburns: If it's no longer valid to have a start time after your end time (ex. working from 8 pm to 8 am) then we should throw an exception or print some error message

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/agents/ScheduledAgent.kt
jeyrschabu: Think it's different than the groovy last line is the return thing. But for readability, I can see the traditional return idiom to be clearer.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
```diff
@@ -26,9 +26,9 @@ import java.time.LocalTime
 import java.time.ZoneId
 
 /**
- * @param minImagesUsedByLC minimum number of images used by launch configurations that should be
+ * @param {@link #minImagesUsedByLC} minimum number of images used by launch configurations that should be
```
jeyrschabu: quick fix to the doc here

PR #214, opened 2019-06-19
refactor(*): Using a queue for work processing

jeyrschabu:
Currently a swabbie node races to acquire a lock on as many concurrent work items as it can obtain locks on.
As we have started to process more cloud resources, this scheme leads to a few problems:
- A node tries to process too much in a cycle, leading to degraded performance, high & bursty memory footprint requiring instance types with more RAM
- Hard to troubleshoot when more than one resource is being processed. (Log chattiness does not help either)
- Resource handlers running on their own threads adds unneeded complexity

This PR aims at simplifying how work is processed by swabbie. The new flow is a loose Producer/Consumer pattern 
where a `WorkQueueManager` adds work to the `WorkQueue` to be processed by handlers. 
Nodes consume work present on the queue one at a time per cycle until all work is complete. 
The `WorkQueueManager` monitors the queue and refills it when empty. This eliminates the need for a thread per handler,
improves efficiency and performance of worker nodes, removes the complexity of concurrency and post action callbacks.

Work is more fairly distributed amongst worker nodes as result.


emjburns:
(Porting from slack) 

Current state:
- Swabbie goes through every work item (workConfiguration) and dispatches that to handlers
- Once a handler gets the work item, a node picks the work, ensures it's the only one working on it using a lock
   * The node claims the work if it can successfuly acquire a lock for the work
   * The node that starts first will mostly have priority of work
- Handlers are assigned multiple work items in a single cycle
   * This may lead some nodes to do a little too much and may starve other resources i.e GC cycles ect...

Simple queue  state:
- Swabbie will initialize work items (workConfiguration) and place them on a queue
- A node will pick the work off the queue and ensure it's the only one working on it using a lock (just like current state)
   * This ensures that one work item is being worked on by a node in a cycle
- Once the queue of work is empty, Swabbie will refill it

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -516,78 +469,75 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
...
+  private fun doDelete(workConfiguration: WorkConfiguration) {
     exclusionCounters[Action.DELETE] = AtomicInteger(0)
     val candidateCounter = AtomicInteger(0)
     val markedResources = mutableListOf<MarkedResource>()
-    try {
-      val currentMarkedResourcesToDelete = resourceRepository.getMarkedResourcesToDelete()
-        .filter {
-          it.notificationInfo != null && !workConfiguration.dryRun && it.namespace == workConfiguration.namespace
-        }
-        .sortedBy { it.resource.createTs }
```
emjburns: Why did you remove the `try / catch` from here (and notify)? Do you think that we won't encounter any errors? Or are they handled someplace else?

jeyrschabu: Removed the `try ... finally` that was there  to ensure the `postAction` callbacks get called.
Those callbacks are no longer needed. There is a try/catch to handle exception when an individual resource causes to throw.

Other exceptions that bubble up are handled by the caller: `ScheduledAgent#process` https://github.com/spinnaker/swabbie/pull/214/files#diff-facdb46f9906b5a7113a4110c7cc1813L146

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -516,78 +469,75 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
...
-
-      printResult(
-        candidateCounter,
-        AtomicInteger(currentMarkedResourcesToDelete.size),
-        workConfiguration,
-        markedResources,
-        Action.DELETE
-      )
-    } finally {
-      postClean.invoke()
```
aravindmd: why do we not want any of these post actions? anymore

jeyrschabu: They aren't needed. They used to be hooks to let the caller know the task is done to execute something after. This was a side effect of using different threads before. I have been meaning to review these for a bit now.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/agents/ScheduledAgent.kt
```diff
@@ -128,23 +91,31 @@ abstract class ScheduledAgent(
...
-      when (action) {
-        Action.MARK -> it.mark(workConfiguration, onCompleteCallback)
-        Action.NOTIFY -> it.notify(workConfiguration, onCompleteCallback)
-        Action.DELETE -> it.delete(workConfiguration, onCompleteCallback)
-        else -> log.warn("Unknown action {}", action.name)
+  /**
+   * Takes work [WorkConfiguration] off the [WorkQueue], acquires a lock and
+   * dispatches the work to a suitable handler for an [Action] for processing
+   */
+  override fun process() {
```
emjburns: The work queue is loaded with the work configuration, which is `provider:account:region:resourceType`. This means that if snapshots and images are both enabled for `aws:test:us-east-1` then we have two work configuration entries:
[`aws:test:us-east-1:image`,
 `aws:test:us-east-1:snapshot`]

This function is scheduled for every agent (mark/notify/delete) every `intervalSeconds`. But, since the work queue is only size 2, and gets refilled only on a fixed delay (`swabbie.queue.monitorIntervalMs`), we do a lot less work than I'd expect. (2 operations instead of all 6 operations - mark notify delete for each work configuration - in the interval)

I would like for this work queue to have each type of work in it (mark/notify/delete for each work configuration) and then each instance picks this up one at a time, and the queue gets refilled after `intervalSeconds` only if the queue is empty.

I'd expect swabbie to work like this:
- work queue is populated with each action for each work configuration, which would look like
[`MARK` `aws:test:us-east-1:image`,
 `MARK` `aws:test:us-east-1:snapshot`,
 `NOTIFY` `aws:test:us-east-1:image`,
 `NOTIFY` `aws:test:us-east-1:snapshot`
 `DELETE` `aws:test:us-east-1:image`,
 `DELETE` `aws:test:us-east-1:snapshot` ]
- each instance picks one item of work off the queue, and does it.
- when that instance is done with that item, it picks another item off the queue
- when the queue is empty, the instance does nothing
- every `intervalSeconds` (for each agent type?) the queue is refilled with those work items

This way, all the work is done each interval as we would expect, but each instance does only one thing at a time.

jeyrschabu: Missed that, agreed should be 6 work items. Will update that part.

So is your desire to see an instance to continuously do work as long as the queue has work?
I think that's reasonable. @emjburns

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
emjburns: 👍

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/agents/ScheduledAgent.kt
emjburns: Ya, I think that makes sense behavior-wise, so all the work gets done. Then we don't have to deal with scheduling things!

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkProcessor.kt
```diff
@@ -0,0 +1,108 @@
...
+  private val lockingService: LockingService
+) {
+  private val log: Logger = LoggerFactory.getLogger(javaClass)
+  private val workId = registry.createId("swabbie.work")
+
+  /**
+   * Takes work [WorkItem] off the [WorkQueue], acquires a lock and
+   * dispatches the corresponding work configuration to a suitable handler for processing
+   */
+  @Scheduled(fixedDelayString = "\${swabbie.work.intervalMs:5000}")
```
jeyrschabu: `5000Ms` too tight for a default. Will bump that up

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
```diff
@@ -0,0 +1,144 @@
...
+    get() = {
+      ifEnabled {
+        ensureLoadedCaches()
+        queue.refillOnEmpty()
+      }
+    }
+
+  override val onDiscoveryDownCallback: (event: RemoteStatusChangedEvent) -> Unit
+    get() = {
+      !ifEnabled {
```
emjburns: Shouldn't this always clear the queue? Because, if an instance goes down in discovery we want to make sure it doesn't do more work (even if it's during a scheduled work time)

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
```diff
@@ -0,0 +1,144 @@
...
+      }
+    }
+
+  /**
+   * Monitors and refills the [WorkQueue] if empty.
+   * The queue is emptied if swabbie is disabled via a persisted property/config or outside the configured schedule
+   */
+  @Scheduled(fixedDelayString = "\${swabbie.queue.monitorIntervalMs:5000}")
+  fun monitor() {
+    !ifEnabled {
```
emjburns: nit: this seems overly kotlin-y for what seems to be an if-else statement. My preference is to use a more boring syntax for these scheduling things to make it really clear what is going on

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/work/WorkProcessorTest.kt
```diff
@@ -0,0 +1,211 @@
...
+
+object WorkProcessorTest {
+  @Test
+  fun `should do nothing if work queue is empty`() {
+    val workQueue = InMemoryWorkQueue()
+    val handler = mock<ResourceTypeHandler<TestResource>>()
+
+    whenever(handler.handles(any())) doReturn true
+
+    val processor = WorkProcessor(
```
emjburns: You define this for every test in here. Can we have this in either a common function, or just as the `@Subject` (or the equivalent - i can't remember if thats the right syntax for this test framework, it's probably not) of the class?

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/work/WorkProcessorTest.kt
```diff
@@ -0,0 +1,211 @@
...
+    processor.process()
+
+    verify(handler, never()).notify(any())
+    verify(handler, never()).delete(any())
+    verify(handler, never()).mark(any())
+  }
+
+  @Test
+  fun `should process work present on the queue`() {
+    val workConfiguration1 = WorkConfiguration(
```
emjburns: similarly, let's pull these declarations up to the top so they can be shared! makes the tests more readable

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
```diff
@@ -0,0 +1,144 @@
...
+        // should only empty the queue if disabled.
+        queue.clear()
+      }
+    }
+
+  /**
+   * Monitors and refills the [WorkQueue] if empty.
+   * The queue is emptied if swabbie is disabled via a persisted property/config or outside the configured schedule
+   */
+  @Scheduled(fixedDelayString = "\${swabbie.queue.monitorIntervalMs:5000}")
```
emjburns: This should probably default to a longer time? I would say running the agents once every 15 minutes is a good default time for cleaning things? Every 5 seconds seems a little quick.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkProcessor.kt
emjburns: Ya I agree, checking once every minute seems good?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkProcessor.kt
```diff
@@ -0,0 +1,108 @@
...
+        } else {
+          process(work)
+        }
+      } catch (e: Exception) {
+        log.error("Error while processing work", e)
+      }
+    } while (!workQueue.isEmpty())
+  }
+
+  private fun process(work: WorkItem) {
```
emjburns: I love how this is now simplified a lot! I think it's easier to understand and reason about. Nice work.

emjburns:
I love the direction this is going! This reduces complexity, and is easier to reason about. Also, it stops swabbie-the-overachiever from doing too much at once and becoming unhealthy. 

A few small comments, but overall looks great.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
jeyrschabu: If we were running one instance yes. But since the queue is shared, we should probably not empty it since other instances UP might be using it.

On a second thought, perhaps we don't even need a redisQueue at all. An internal InMemoryQueue would work fine, especially with the lockManager. That way each instance can manage their in memory queue. I actually like this better.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkProcessor.kt
jeyrschabu: A min a two. Definitely not 5sec for a default.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
jeyrschabu: Scratch that. We need it lol

jeyrschabu: And we should only clear the queue if disabled. So this should stay as is aside for the style update

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisWorkQueue.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.netflix.spinnaker.swabbie.work.WorkQueue
+import org.springframework.stereotype.Component
+
+/**
+ * Stores an list of [com.netflix.spinnaker.swabbie.model.WorkItem]
+ * @property _seed work configurations defined in the application config from which to derive work items
+ * @see [com.netflix.spinnaker.swabbie.work.WorkQueueManager]
+ */
+@Component
+class RedisWorkQueue(
```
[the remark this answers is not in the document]
asher: The current implementation will result in the queue containing N duplicate workItems when running N swabbie instances, since each will call `seed()` without restriction on what gets pushed onto the queue. Using a redis set that is added to via `sadd` and pulled from via `spop` would solve this. Should also remove the need to shuffle the workItems, since `spop` read order should be pseudo-random.

PR #217, opened 2019-06-27
fix(metrics): Fix delete metrics

aravindmd:
- We were not publishing `DeleteResourceEvent` for delete action , adding that to take care of metrics issue.


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -752,6 +753,12 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
     }
   }
 
+  private fun sendDeleteEvents(
```
emjburns: nit: this probably doesn't need to be a function because it's already one line: 
`resources.forEach { applicationEventPublisher.publishEvent(DeleteResourceEvent(it, workConfiguration))`

aravindmd: updated

PR #227, opened 2019-07-15
Entity tag fix

aravindmd:
- Pass `id` as part of `orca` request that is required to fetch tags from `front50` during `DeleteEntityTagsTask`

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/tagging/EntityTag.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/tagging/EntityTag.kt
@@ -16,13 +16,15 @@
 
 package com.netflix.spinnaker.swabbie.tagging
 
+const val SWABBIE_ENTITY_TAG_NAME = "spinnaker_ui_alert:swabbie_deletion_candidate"
+
 data class EntityTag(
 
   val namespace: String,
   val value: TagValue?,
   val valueType: String = "object",
   val category: String = "Clean Up",
-  val name: String = SwabbieEntityTag.swabbieEntityTag
+  val name: String = SWABBIE_ENTITY_TAG_NAME
 ) : Tag
 
 data class TagValue(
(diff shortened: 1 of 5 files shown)


swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/tagging/OrcaTaggingService.kt
```diff
@@ -44,7 +44,8 @@ class OrcaTaggingService(
               OrcaJob(
                 type = tagRequest.type,
                 context = mutableMapOf(
-                  "tags" to listOf(tagRequest.id)
+                  "tags" to listOf(tagRequest.id),
+                  "id" to tagRequest.id
```
emjburns: So `id` needs to be the entity id, and `tags` is a list of strings of the tag id. These shouldn't be the same value I think?

aravindmd: fixed it to pass the correct tag and id , tested it locally and looks good now

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/tagging/EntityTag.kt
```diff
@@ -55,3 +56,9 @@ open class EntityTagRequest(
   open val application: String,
   open val description: String
 ) : TagRequest
+
+open class SwabbieEntityTag {
```
emjburns: you can probably just do 
`const val SWABBIE_ENTITY_TAG_NAME = "spinnaker_ui_alert:swabbie_deletion_candidate"`
in this file. What is the use of the class with a companion object?

aravindmd: i added it as a class with an companion object to access it in swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/tagging/OrcaTaggingService.kt

aravindmd: nvm , fixed it

PR #229, opened 2019-07-18
refactor(*): cleaning up aws api

jeyrschabu:
- Refactor edda bits
- cleanup to have a common interface for aws
- fix a few behavior around queue processor to wait for caches to load


swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/edda/Edda.kt
```diff
@@ -0,0 +1,412 @@
...
+class Edda(
+  eddaApiClients: List<EddaApiClient>,
+  private val retrySupport: RetrySupport,
+  private val registry: Registry
+) : AWS, EddaApiSupport(eddaApiClients, registry) {
+  private val log: Logger = LoggerFactory.getLogger(javaClass)
+
+  override fun getLaunchConfigurations(params: Parameters): List<AmazonLaunchConfiguration> {
+    withEddaClient(
+      region = params.region,
```
emjburns: can we make it so `withEddaClient` takes a `Parameters` object instead of duplicating this over and over?

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/config/EddaConfiguration.kt
```diff
@@ -100,26 +107,43 @@ open class EddaConfiguration {
     return listOf(accountEddas, endpointEddas).flatten()
   }
 
+  @Bean
+  open fun eddaEndpointCache(eddaEndpointsService: EddaEndpointsService): EddaEndpointCache {
+    return EddaEndpointCache(eddaEndpointsService)
```
emjburns: In a future step, these will default to just vanilla aws unless `edda.enabled: true` is set, right?

jeyrschabu: Yeah. I am planning to rev this up once I put in aws stuff.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/edda/Edda.kt
jeyrschabu: Yeah, this class needs to be refactored. Needs to be cleaned up. I will take of this in this PR

jeyrschabu: Going to tackle this in another PR as discussed

PR #231, opened 2019-07-22
Fix ASG tagging during optout flow

aravindmd:
- Fix ASG tagging during optout flow
- update the `tagResource` function to handle `image` and `serverGroup` `resourceType`


aravindmd:
Going to test this locally for a few resources

aravindmd:
> Would love to see this lifted in a `ResourceTagger` (different than the Entity Tag one). That class would then be responsible for applying the tag based on the resource type.

Discussed with @jeyrschabu and agree that this needs to be refactored, but will tackle that next in a different PR, for now we agreed to go with this approach. cc @emjburns

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceStateManager.kt
```diff
@@ -171,28 +172,51 @@ class ResourceStateManager(
...
     workConfiguration: WorkConfiguration
   ): String {
-    val taskId = taggingService.upsertImageTag(
-      UpsertImageTagsRequest(
-        imageNames = setOf(resource.name ?: resource.resourceId),
+    val taskId = with(resource.resourceType) {
+      when {
+        equals("serverGroup", true) -> tagAsg(resource)
+        equals("image", true) -> tagImage(resource)
+        else -> log.error("Failed to tag resource ${resource.uniqueId()}")
```
emjburns: When this is the case (not server group or image) `taskId` will be null and then it will NPE below, right?

Also, what happens in the case of a snapshot?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/tagging/AmazonServerGroupTag.kt
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.swabbie.tagging
+
+open class AmazonServerGroupTagsRequest(
```
emjburns: nit: doesn't seem like we need to have these two classes, could they be combined into one class?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceStateManager.kt
aravindmd: good point , yeah we don't have an option to optout today for any other resource types today, do you think it would be better to handle this by throwing an exception instead ?

emjburns: We should have the option to opt out for snapshot - and I thought that was working.

If we correctly opt out the resource but just don't know how to tag it then I don't think we should throw an exception here, but we will need to handle a `null` `taskId` below.

aravindmd: Updated the code for this. You are right `snapshot` outputs work fine , i was referring to the `aws` `expiration_time` tagging.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/tagging/AmazonServerGroupTag.kt
aravindmd: done

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceStateManager.kt
```diff
@@ -176,19 +176,19 @@ class ResourceStateManager(
       when {
         equals("serverGroup", true) -> tagAsg(resource, workConfiguration)
         equals("image", true) -> tagImage(resource)
-        else -> log.error("Failed to tag resource ${resource.uniqueId()}")
+        else -> log.error("Failed to tag resource ${resource.uniqueId()} with aws expiration_time tag ")
       }
     }
     taskTrackingRepository.add(
-      taskId.toString(),
+      taskId!!.toString(),
```
emjburns: The `!!` asserts that this will never be `null`. Instead, you need to correctly handle if the `taskId` is `null` by not adding it to the `taskTrackingRepository`.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/tagging/AwsTag.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Netflix, Inc.
+ * Copyright 2018 Netflix, Inc.
```
emjburns: It's 2019

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceStateManager.kt
aravindmd: done

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceStateManager.kt
```diff
@@ -162,37 +163,68 @@ class ResourceStateManager(
...
     workConfiguration: WorkConfiguration
-  ): String {
-    val taskId = taggingService.upsertImageTag(
-      UpsertImageTagsRequest(
-        imageNames = setOf(resource.name ?: resource.resourceId),
+  ): String? {
+    var taskId: String? = with(resource.resourceType) {
+      when {
+        equals("serverGroup", true) -> tagAsg(resource, workConfiguration)
+        equals("image", true) -> tagImage(resource)
```
emjburns: What about `snapshot`?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceStateManager.kt
```diff
@@ -162,37 +163,68 @@ class ResourceStateManager(
...
+        else -> {
+          log.error("Failed to tag resource ${resource.uniqueId()} with aws expiration_time tag ")
```
emjburns: This is not the right message - it's that we have no _way_ to tag the resource because we don't know how, or it's not supported. I'd change to something like:

```
log.warn("Resource type $it cannot be tagged with an infrastructure tag.") 
```
Or something like that. (not sure if the `$it` syntax is right when accessing that resource type).

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceStateManager.kt
```diff
@@ -162,37 +163,68 @@ class ResourceStateManager(
     if (event is OptOutResourceEvent) {
       log.debug("Tagging resource ${event.markedResource.uniqueId()} with \"expiration_time\":\"never\"")
```
emjburns: we could probably pull both these logs into the `tagResource(..)` function, and we wouldn't need to return a value from the function. That would make the `tagResource(..)` function a bit more readable.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceStateManager.kt
aravindmd: Discussed offline with @emjburns , added a TODO for now , needs updates in `orca` and `clouddriver`

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceStateManager.kt
aravindmd: updated with a better log message

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceStateManager.kt
aravindmd: updated `tagResource` function to handle these comments

PR #232, opened 2019-07-22
feat(aws): Adding aws api support

jeyrschabu:
- added support for reading aws directly
- support for loading aws credentials via a bastion configuration
Additional:
- make image handler depend on needed caches
- make image handler caches optional based on if image handling is turned on
- make data classes compatible between edda & vanilla


swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandlerTest.kt
```diff
@@ -1,498 +1,498 @@
...
-
-    return WorkConfigurator(
-      swabbieProperties = swabbieProperties,
-      accountProvider = accountProvider,
-      exclusionPolicies = listOf(AccountExclusionPolicy()),
-      exclusionsSuppliers = Optional.empty()
-    ).generateWorkConfigurations()[0]
-  }
-}
+// /*
```
emjburns: I'm guessing you don't want this whole class to be commented out?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Account.kt
```diff
@@ -46,7 +46,9 @@ data class SpinnakerAccount(
   override val edda: String?,
   override val regions: List<Region>?,
   override val environment: String,
-  override val grouping: Grouping? = null
+  override val grouping: Grouping? = null,
+  val assumeRole: String = "role/spinnaker",
```
emjburns: should this be configurable?

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/AWS.kt
```diff
@@ -40,3 +88,398 @@ interface AWS {
...
+  private val log: Logger = LoggerFactory.getLogger(javaClass)
+
+  init {
+    if (proxy.isPresent) {
+      log.info("Using aws proxy {}", proxy.javaClass.name)
+    }
+  }
+
+  override fun getInstances(params: Parameters): List<AmazonInstance> {
+    if (proxy.isPresent) {
```
emjburns: I don't love that every function here has this `if (proxy.isPresent) {` bit which does the same thing. Since it looks like if there _is_ a proxy it is applicable for every call, can we make this cleaner? We can discuss offline on ways to make this nicer.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Account.kt
jeyrschabu: Yeah, will do that in another change.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/AWS.kt
```diff
@@ -40,3 +87,326 @@ interface AWS {
...
+  }
+
+  override fun getSnapshot(params: Parameters): AmazonSnapshot? {
+    val account: SpinnakerAccount = findAccount(params) ?: return null
+    val request = DescribeSnapshotsRequest().withSnapshotIds(params.id)
+    val result = account.ec2(params.region).describeSnapshots(request)
+    return convert<AmazonSnapshot>(result.snapshots).first()
+  }
+
+  private val snapshotsMaxResult = 1000
```
aravindmd: should these values come from a config ?

jeyrschabu: same as the other constants. Will address in another changeset

PR #233, opened 2019-07-23
feat(rule): Adding a general tag rule

jeyrschabu:
- added  `ExpiredResourceRule` to apply to resources that are tagged with a ttl.
- rule evaluates if resource has expired.


jeyrschabu:
Works if this is on the resource type config:
```rule:
            attributes:
            - key: ownedBy
              value:
                - pattern:^packer-stuff```

**Comment by aravindmd** [comment]
Nice ! , do you think adding a `@conditional` or something else like a kill switch to safeguard this would make sense ?

File: swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/TemporalThresholdRule.kt [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+import com.netflix.spinnaker.swabbie.model.Summary
+import com.netflix.spinnaker.swabbie.model.WorkConfiguration
+import org.springframework.stereotype.Component
+
+/**
+ * A general rule that applies when temporal tags are found on an amazon resource
+ * When the resolved resource age is older than today, the resource will be marked for deletion if
+ * [WorkConfiguration.rule] attribues match the resource
+ */
+@Component
```
emjburns: I find this confusing at first glance. Can you give an example with a tag and the work configuration and when it would be deleted here in this javadoc?

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/exclusions/AmazonTagExclusionPolicy.kt
```diff
@@ -25,46 +25,24 @@ import org.springframework.stereotype.Component
...
-                            return null
-                          else -> {
-                            // no need to check age here.
-                            log.debug("Resource age comparison with {}. Result: {}", excludable.createTs, it)
-                          }
-                        }
-                    }
-                  }
-                }
+    if (excludable !is AmazonResource || excludable.tags().isNullOrEmpty()) {
```
emjburns: this refactor is delightful, thank you!

swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
```diff
@@ -183,6 +184,14 @@ class Attribute {
   }
 }
 
+/**
+ * A simple configured rule
+ */
```
emjburns: For some reason I'm having trouble understanding this based on the name. Is `BasicRule` just a key/value match rule?

swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/TemporalThresholdRuleTest.kt
```diff
@@ -0,0 +1,79 @@
...
+      name = "tags",
+      value = listOf(
+        mapOf("fooKey" to "fooValue", "ttl" to "1d")
+      )
+    ) as AmazonAutoScalingGroup
+
+    val result = TemporalThresholdRule<AmazonAutoScalingGroup>(listOf(configuration)).apply(asg)
+    Assertions.assertNotNull(result.summary)
+  }
+}
```
aravindmd: would be good to add a test for a non amazon resource type

aravindmd:
> Overall this looks great to me. My one request always is to include more verbose class comments that explain how to use something or give an example (I've added one comment in the specific area on this PR).
> 
> For things that are exposed via config, I find it _super_ helpful to have an example in the class comments for when I'm trying to easily understand how to enable a feature. I think this is arguably worse than actual documentation, but it is easy and more likely to get changed as functionality changes 😆

+ 1 ,  the comments of test controller are really helpful.

jeyrschabu:
Keeping it simple. `BasicRule` was a bit over engineered, so removing that for now and keeping the rule to apply when a resource has expired. The expiration time is determined by inspecting the tags on the resource.

PR #234, opened 2019-08-05
Fix exception while parsing ASG creationDate

aravindmd:


emjburns:
looks like weird git history here?

aravindmd:
> looks like weird git history here?

Yes , noticed the same , although i did a pull from upstream master before the branch, let me squash the commits

PR #236, opened 2019-08-07
fix(tasks): add proper user identification

emjburns:
Fix for the tasks view showing `User: null (swabbie@spinnaker.io)`

### DIFF
--- a/swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaService.kt
+++ b/swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaService.kt
@@ -73,10 +73,15 @@ class OrchestrationRequest(
   val job: List<OrcaJob>
 )
 
+// todo emjburns: service account should be configurable
 class OrcaJob(
   type: String,
   context: MutableMap<String, Any?>
-) : HashMap<String, Any?>(context.apply { put("type", type) }
+) : HashMap<String, Any?>(
+  context.apply {
+    put("type", type)
+    put("user", "swabbie@spinnaker.io")
+  }
 )
 
 /**


swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaService.kt
```diff
@@ -73,10 +73,15 @@ class OrchestrationRequest(
   val job: List<OrcaJob>
 )
 
+// todo emjburns: service account should be configurable
```
ajordens: The user specified here isn't really the `service account` (although that must be specified somewhere else under the covers).

I don't think there's harm in this but the UI should really avoid `User: null` when it has authenticated details it could just show instead.

emjburns: We specify it here: https://github.com/spinnaker/swabbie/blob/master/swabbie-retrofit/src/main/kotlin/com/netflix/spinnaker/swabbie/retrofit/SwabbieRetrofitConfiguration.kt#L51. Service account is probably the wrong phrase.

The ui shows this when the user passed (in this case there wasn't one) doesn't match the actual credentials. Adding it into the payload lets the ui just show the "email". Do you think there's harm in doing this/merging this PR?

emjburns: cc @alanmquach

alanmquach: > I don't think there's harm in this but the UI should really avoid User: null when it has authenticated details it could just show instead.
I can do this.

alanmquach: Done: https://github.com/spinnaker/deck/pull/7305

emjburns:
Closed in favor of the deck pr

PR #240, opened 2019-08-09
fix(serialization): add aws object mapper, change startTime format

emjburns:
This PR does several things.

- Add awsObjectMapper so that we read aws objets more like clouddriver in the case of edda or vanilla aws.
- Adds logs.
- Updates `AmazonSnapshot` to handle either a date or time as `startTime`.
- Deletes the outdated `aws-edda` package. There was nothing there.
- Makes sure we cache instances in regions that are `deprecated` in clouddriver. This status doesn't have a meaning other than UI guidance and we should still look at those instances.


swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/snapshots/AmazonSnapshot.kt
```diff
@@ -41,13 +43,29 @@ data class AmazonSnapshot(
...
   override val name: String = snapshotId,
-  private val creationDate: String? = Dates.toCreationDate(startTime)
+  creationDate: String? = getCreationDate(startTime)
 ) : AmazonResource(creationDate) {
-  override fun equals(other: Any?): Boolean {
-    return super.equals(other)
-  }
+  val startTime: Long = convertStartTime(startTime)
+
+  /**
```
emjburns: This is the big fix of this PR.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/config/AwsConfiguration.kt
```diff
@@ -44,6 +47,12 @@ import java.time.Clock
 open class AwsConfiguration {
   private val defaultRegion = "us-west-2" // TODO: (Jeyrs) Make configurable
 
+  // AWS object mapper ensures edda and vanilla aws responses are the same
+  @Bean
+  @Qualifier("amazonObjectMapper")
```
robzienert: You don't need `@Qualifier`, as the name of the factory method is the same value.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/caches/ImagesUsedByInstancesProvider.kt
```diff
@@ -37,7 +37,9 @@ class ImagesUsedByInstancesProvider(
     accountProvider.getAccounts()
       .filter { it.cloudProvider == "aws" && !it.regions.isNullOrEmpty() }
       .forEach { account ->
-        account.regions!!.filterNot { it.deprecated }.forEach { region ->
+        account.regions!!.forEach { region ->
+          // we need to read all the instances in every region, no matter if it's depricated or not,
```
robzienert:
```suggestion
          // we need to read all the instances in every region, no matter if it's deprecated or not,
```

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/instances/AmazonInstance.kt
```diff
@@ -20,21 +20,22 @@ import com.fasterxml.jackson.annotation.JsonTypeName
...
   val imageId: String,
   val tags: List<Map<String, String>>,
-  private val launchTime: Date,
+  private val launchTime: Long,
   override val resourceId: String = instanceId,
   override val resourceType: String = INSTANCE,
   override val cloudProvider: String = AWS,
   override val name: String = instanceId,
-  private val creationDate: String? = LocalDateTime.ofInstant(launchTime.toInstant(), ZoneId.systemDefault()).toString()
+  private val creationDate: String? =
```
robzienert: Does this need to be `private val` since it's just being passed to the super constructor?

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/snapshots/AmazonSnapshot.kt
```diff
@@ -41,13 +43,29 @@ data class AmazonSnapshot(
...
+   * Amazon returns [startTime] as a String (date), whereas edda returns it as a Long.
+   * This allows either of those formats to be accepted and used successfully.
+   */
+  companion object {
+    private fun convertStartTime(value: Any): Long {
+      return when (value) {
+        is String -> Dates.toLocalDateTime(value).toInstant(ZoneId.systemDefault().rules.getOffset(Instant.now())).toEpochMilli()
```
robzienert: Let's line-wrappy this.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/instances/AmazonInstance.kt
emjburns: You can't do passthrough values with data classes apparently

PR #241, opened 2019-08-12
Clean state repository

emjburns:
We use the resource state repository to track when resources are opted out/marked/notified/deleted. Right now we never remove any resources from it. We've deleted a lot of things, and it would be nice to remove them so that we don't have to deal with weird old data.

This PR adds a cleaning function that runs every 5 minutes to remove deleted resources from the resource state repository.


swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceStateRepository.kt
```diff
@@ -112,6 +119,17 @@ class RedisResourceStateRepository(
     }
   }
 
+  override fun remove(resourceId: String, namespace: String) {
+    "$namespace:$resourceId".let { id ->
+      statesKey(id).let { key ->
```
robzienert: Doesn't look like `key` is used anywhere here? Are we supposed to be using `key` in the redis commands instead?

emjburns: good call, we are not. removing.

PR #243, opened 2019-08-13
Rest when disabled

emjburns:
Swabbie was working when it was disabled in discovery. This PR fixes that functionality so that swabbie doesn't work when it's down in discovery, and also adds a test for that case.

I also added a testing util and refactored some of the tests.

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/WorkConfiguration.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/WorkConfiguration.kt
@@ -43,7 +43,7 @@ data class WorkConfiguration(
 ) {
   private val log = LoggerFactory.getLogger(javaClass)
 
-  fun toLog(): String = "$namespace/$account/$location/$cloudProvider/$resourceType"
+  fun toLog(): String = "$namespace/${account.accountId}/$location/$cloudProvider/$resourceType"
   fun toWorkItems(): List<WorkItem> {
     log.info("Actions $enabledActions enabled for ${toLog()}")
     return enabledActions.map {
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/tagging/ResourceEntityTagger.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/tagging/ResourceEntityTagger.kt
@@ -60,10 +60,11 @@ class ResourceEntityTagger(
   }
 
   private fun tagMessage(markedResource: MarkedResource, workConfiguration: WorkConfiguration): String {
-    return markedResource.summaries
+    val summaries = markedResource.summaries
       .joinToString(", ") {
         it.description
-      }.also { summary -> return formatMessage(summary, markedResource, workConfiguration) }
+      }
+    return formatMessage(summaries, markedResource, workConfiguration)
   }
 
   private fun formatMessage(message: String, markedResource: MarkedResource, workConfiguration: WorkConfiguration): String {
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
@@ -34,6 +34,7 @@ import java.time.LocalDate
 import java.time.LocalDateTime
 import java.time.LocalTime
 import java.time.Duration
+import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * Serves as the producer of work for resource handlers, monitors and refills the work queue once all work is complete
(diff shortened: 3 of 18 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/WorkConfiguration.kt
```diff
@@ -43,7 +43,7 @@ data class WorkConfiguration(
 ) {
   private val log = LoggerFactory.getLogger(javaClass)
 
-  fun toLog(): String = "$namespace/$account/$location/$cloudProvider/$resourceType"
+  fun toLog(): String = "$namespace/${account.accountId}/$location/$cloudProvider/$resourceType"
```
emjburns: `account` got changed to an object. Grabbing the id here for a nice log.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/tagging/ResourceEntityTagger.kt
```diff
@@ -60,10 +60,11 @@ class ResourceEntityTagger(
   }
 
   private fun tagMessage(markedResource: MarkedResource, workConfiguration: WorkConfiguration): String {
-    return markedResource.summaries
+    val summaries = markedResource.summaries
```
emjburns: intellij was telling me there was unreachable code here. Refactoring to be less kotlin-y

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
```diff
@@ -84,7 +86,16 @@ class WorkQueueManager(
     }
   }
 
-  private fun isEnabled(): Boolean {
+  /**
+   * Swabbie is considered disabled if it is out of service in discovery, disabled via fast property, or the
+   *  schedule determines it is off hours.
+   */
+  internal fun isEnabled(): Boolean {
```
emjburns: Added discovery configuration to this function to make sure swabbie wouldn't work while down.

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/ResourceTypeHandlerTest.kt
```diff
@@ -736,34 +732,4 @@ object ResourceTypeHandlerTest {
     verify(notifier, times(0)).notify(any(), any(), any())
     verify(resourceRepository, times(1)).upsert(updatedMarkedResource)
   }
-
-  internal fun workConfiguration(
```
emjburns: pulled this to a common util, `WorkConfigurationTestHelper.generateWorkConfiguration()`

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
```diff
@@ -65,10 +69,8 @@ class WorkQueueManager(
 
   override val onDiscoveryDownCallback: (event: RemoteStatusChangedEvent) -> Unit
     get() = {
-      if (!isEnabled()) {
-        // should only empty the queue if disabled.
-        queue.clear()
-      }
+      up.set(false)
+      queue.clear()
```
asher: I don't think this is the right behavior here. With `RedisWorkQueue`, queue state is shared across instances. If we're running 3 swabbie instances and one is disabled, that shouldn't impact the other 2.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
```diff
@@ -84,7 +86,16 @@ class WorkQueueManager(
     }
   }
 
-  private fun isEnabled(): Boolean {
+  /**
+   * Swabbie is considered disabled if it is out of service in discovery, disabled via fast property, or the
+   *  schedule determines it is off hours.
+   */
+  internal fun isEnabled(): Boolean {
+    if (!up.get()) {
```
asher: This also means disabling one instance in discovery stops healthy and UP instances due to the else condition in `monitor`. Another approach that avoids this sort of problem would be to only gate `WorkProcessor.process()` on discovery state.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
emjburns: 😌 oh crap you are totally right. I shall fix this.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
emjburns: Great catch. Thanks. I'm going to make the work processor also discovery activated so that it only processes when it's up.

PR #258, opened 2019-10-07
refactor(notifications): Updating notification logic

jeyrschabu:
- remove notificationConfiguration.required field
- remove notificationCount
- added a notification queue
- misc refactor for items no longer fitting the overall architecture


jeyrschabu:
Notifications as currently designed would result in:

- Published notification events for resources not being notified on
- Too frequent notifications for small sets of resources to  an owner
- Separate notifications to the same user for different namespaces i.e `aws:servergroup:test:us-east-1` vs `aws:servergroup:test:us-west-2`

This PR solves the aforementioned problems by:
- Ensuring no events is published for resources not being notified on (amis & snapshots)
- Scheduling notifications to ensure more resources are packaged a notification to be sent to an owner
- Grouping notifications by resource type, ensuring resources belonging to different account/region pairs can be contained in the same notification

swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
```diff
@@ -69,25 +69,11 @@ class CloudProviderConfiguration {
 }
 
 /**
- * Notifications have both an [enabled] and a [required] configuration.
```
jeyrschabu: No longer applicable

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -108,10 +102,50 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
   }
 
   /**
-   * Notifies resource owners
+   * Notifies for everything in resourceRepository.getMarkedResources() that doesn't have notificationInfo
    */
   override fun notify(workConfiguration: WorkConfiguration) {
```
jeyrschabu: This now schedules notifications when enabled, otherwise update the resource notification info so it can be ready for deletion. AMIs and Snapshots would fall in this category

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -582,122 +615,46 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
...
+      }
+    val maxItemsToProcess = min(size, getMaxItemsProcessedPerCycle(workConfiguration))
+    return subList(0, maxItemsToProcess)
       .filter {
         @Suppress("UNCHECKED_CAST")
         !shouldExcludeResource(it.resource as T, workConfiguration, optedOutResourceStates, Action.NOTIFY)
       }
-      .groupBy { it.grouping?.value }
-
-    groupedResourcesToNotify.forEach { grouping, resources ->
```
jeyrschabu: Moved relevant bits `EchoNotifier`.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/discovery/DiscoveryAware.kt
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.swabbie.discovery
+
+import com.netflix.spinnaker.kork.eureka.RemoteStatusChangedEvent
+import java.util.concurrent.atomic.AtomicBoolean
+
+open class DiscoveryAware : DiscoveryActivated {
```
jeyrschabu: We have a few places where this is needed. Lifting this in a class so it can be used by any scheduler.

swabbie-echo/src/main/kotlin/com/netflix/spinnaker/swabbie/echo/EchoNotifier.kt
```diff
@@ -16,23 +16,78 @@
...
+
+    throw UnsupportedOperationException("Notification Type not supported in $workConfiguration")
+  }
+
+  private fun String.unCamelCase(): String =
+    split("(?=[A-Z])".toRegex()).joinToString(" ").toLowerCase()
+
+  private fun notificationContext(
+    recipient: String,
+    resources: List<Pair<MarkedResource, WorkConfiguration>>,
```
jeyrschabu: This will require the echo notification template for swabbie to be updated to use this model. It makes determining metadata/configuration associated with a marked resource. We were previously passing just one because notifications were scoped to a specific namespace.  (cc: @aravindmd )

swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
emjburns: Thank goodness 😆

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/discovery/DiscoveryAware.kt
emjburns: What if we just added this up/down feature to `DiscoveryActivated`? I think that any class that is `DiscoveryActivated` wouldn't mind having this up atomic boolean, and that avoids another class to pick from.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
```diff
@@ -0,0 +1,165 @@
...
+
+  /**
+   * @return Marked resources to notify on
+   */
+  private fun groupRelatedResourcesThenNotify(markedResources: List<MarkedResource>) {
+    val now = Instant.now(clock)
+    val tasks = mutableListOf<NotificationTask>()
+    do {
+      val notificationTask: NotificationTask? = notificationQueue.pop()
+      if (notificationTask == null) {
```
emjburns: Could we just check `notificationQueue.isEmpty()` here?

emjburns: I guess, what's the case where the condition in this `do/while` loop would let this run but then there wouldn't be anything to notify? Are we locking on something else that would allow this to happen?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
```diff
@@ -0,0 +1,165 @@
...
+      return
+    }
+
+    lockingService.acquireLock(lockOptions) {
+      val resources = resourceTrackingRepository.getMarkedResources()
+        .filter {
+          it.notificationInfo == null
+        }
+
+      groupRelatedResourcesThenNotify(resources)
```
emjburns: I'd prefer to split this function out into several functions, but you don't have to do that. In general, I prefer things that are structured like: 
```
bigClassMethod() {
  val resultOne = stepOne()
  val resultTwo = stepTwo(resultOne)
  stepThree(resultTwo)
}
```
I find this easier to figure out where the work is in the class, and avoids the problem of "oh this method looks harmless ... three levels down.. oh wait we're doing a lot in here".

When looking at this function I would expect that first resources are grouped (in one method), and then those notifications are sent (in another method). That should be easier to debug and test as well.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
```diff
@@ -69,25 +69,11 @@ class CloudProviderConfiguration {
...
- *
- *  TODO EB: Remove the hard dependency on notifications for all resources.
- *   Instead, make this configurable in a logical and flexible way.
+ *  [enabled] on/off flag for notifications
  */
 open class NotificationConfiguration(
   var enabled: Boolean = false,
-  var required: Boolean = true,
-  var types: MutableList<String> = mutableListOf("Email"),
+  var types: List<String> = listOf("Email"),
```
emjburns: nit: it's unusual that email is capitalized. I'd change to lowercase

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationQueue.kt
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.swabbie.notifications
+
+/**
+ * A queue for storing notification tasks
+ */
+interface NotificationQueue {
+  fun add(notificationTask: NotificationSender.NotificationTask)
```
emjburns: I like it when you add an import for the class, so that it shows up as `notificationTask: NotificationTask`. `option + click` on `NotificationSender` in intellij should bring up that option. It's a little cleaner to read.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
```diff
@@ -0,0 +1,165 @@
...
+        }.groupBy {
+          it.resourceOwner
+        }
+
+      byResourceOwner.forEach { (owner, list) ->
+        val resourcesAndConfigurations: List<Pair<MarkedResource, WorkConfiguration>> = list.toResourceAndConfigurationPairs()
+        val result = notifier.notify(
+          Notifier.Envelope(
+            recipient = owner,
+            resources = resourcesAndConfigurations
```
emjburns: What's the value in passing the whole work configuration here? Does the notification system need everything? If not maybe we could pass a lot less information.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
```diff
@@ -0,0 +1,165 @@
...
+                resourceTrackingRepository.upsert(it)
+                applicationEventPublisher.publishEvent(OwnerNotifiedEvent(it, workConfiguration))
+              }
+
+            registry.counter(notificationsId.withTags("result", "success")).increment()
+          } else {
+            // unmark this resource & let it go through the mark cycle later
+            registry.counter(notificationsId.withTags("result", "failure")).increment()
+            resourceTrackingRepository.remove(resource)
+            applicationEventPublisher.publishEvent(UnMarkResourceEvent(resource, workConfiguration))
```
emjburns: I think this should be a different type of event, so that we can log metrics independently for things that fail to notify. I don't want to lump everything into the "unmark" bucket.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
```diff
@@ -0,0 +1,165 @@
...
+          it.notificationInfo == null
+        }
+
+      groupRelatedResourcesThenNotify(resources)
+    }
+  }
+
+  /**
+   * @return Marked resources to notify on
+   */
```
emjburns: I'd like a method level comment here explaining why we need to group resources and then how we're grouping and notifying. I find this method a little hard to follow.

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/ResourceTypeHandlerTest.kt
```diff
@@ -652,84 +650,73 @@ object ResourceTypeHandlerTest {
...
+
+    whenever(
+      notifier.notify(any())
+    ) doReturn NotificationResult(markedResource.resourceOwner, EMAIL, success = true)
+
+    defaultHandler.setCandidates(mutableListOf(defaultResource))
+    defaultHandler.setRules(defaultRules)
+
     defaultHandler.notify(configuration)
-    verify(notifier, times(1)).notify(any(), any(), any())
```
emjburns: Do you want to verify or assert anything in this test? FYI we're using Strikt in `keel` (http://strikt.io/) and it's really good for asserting simple things. I think I already added it to some of the swabbie tests.

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSenderTest.kt
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  @Test
+  fun `should not notify if there is nothing in the notification queue`() {
+    whenever(
+      resourceRepository.getMarkedResources()
+    ) doReturn listOf(
+      createMarkedResource(workConfiguration = workConfiguration2, id = "3", owner = "test@netflix.com")
+    )
+
```
emjburns: Don't you actually need to call the notifier here?

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSenderTest.kt
```diff
@@ -0,0 +1,167 @@
...
+      verify(notifier, times(1)).notify(capture())
+      val passedResources = firstValue.resources.map { it.first }
+      expectThat(passedResources)
+        .isNotEmpty()
+        .isEqualTo(
+          listOf(resource1, resource2)
+        )
+    }
+
+    // Should publish events for resource1 and resource2 since they match the resource type
```
emjburns: Maybe this test would be better if it was split out into two tests? Like separately test that notification info is populated correctly, and if type doesn't match then the resource is not notified?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
emjburns: Ahh I didn't realize `Envelope` was only going to the `EchoNotifier`. To me envelope has the specific connotation of being used in a request/response body to wrap things. There's probably a clearer name for this.

swabbie-echo/src/main/kotlin/com/netflix/spinnaker/swabbie/echo/EchoNotifier.kt
```diff
@@ -16,23 +16,78 @@
...
+            return NotificationResult(recipient, NotificationType.EMAIL, success = false)
+          }
+        }
+      }
+    }
+
+    throw UnsupportedOperationException("Notification Type not supported in $workConfiguration")
+  }
+
+  private fun String.unCamelCase(): String =
```
emjburns: Is there a commons lang function that already does this?

jeyrschabu: See TODO comment. I'd prefer changing resource types into something that can provide a human readable name rather than trying to normalize this way. Leaving this as-is for now. I will address this in a different refactor.

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSenderTest.kt
jeyrschabu: There are still a couple of scenarios I'd like tested in this class. I will add separate tests.

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/ResourceTypeHandlerTest.kt
jeyrschabu: I will take a look. `Strikt` is used in a few places including in `NotificationSenderTest`

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
jeyrschabu: There is a counter right above tracking notification failures.

Thinking about this more, the resource would get picked up when we notify on the same resource type. So no need to remove and unmark. It could use a debug level log here as well.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
aravindmd: do we have any tests to validate this for AMI's and snapshots ?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/discovery/DiscoveryAware.kt
jeyrschabu: Fine by me, I suppose that can be implemented in `DiscoveryActivated`.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
jeyrschabu: This is good feedback!

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
jeyrschabu: It doesn't have the same connotation to me. I am not married to that name so what's your suggestion?

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSenderTest.kt
jeyrschabu: Planning on revisiting this test

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
jeyrschabu: Please elaborate

swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
jeyrschabu: Whatever fine!

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationQueue.kt
jeyrschabu: if it were a top level class yes. I prefer closer to a fully qualified name for internal classes.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
```diff
@@ -0,0 +1,165 @@
...
+      val byResourceOwner = markedResources
+        .filter {
+          it.resourceType == task.resourceType
+        }.groupBy {
+          it.resourceOwner
+        }
+
+      byResourceOwner.forEach { (owner, list) ->
+        val resourcesAndConfigurations: List<Pair<MarkedResource, WorkConfiguration>> = list.toResourceAndConfigurationPairs()
+        val result = notifier.notify(
```
aravindmd: would we need to check the violations again before sending a notify event ? would there be a case where the violations could have changed when the resource is picked up for notification ?

swabbie-echo/src/main/kotlin/com/netflix/spinnaker/swabbie/echo/EchoNotifier.kt
aravindmd: I will update the template once this is merged.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
jeyrschabu: A `do while` construct allows allows to check for that condition once. Otherwise you would have to check if the queue is empty, then proceed to pop items off of it and continue to check if the queue is empty.

jeyrschabu: Not sure I follow the second part of the comment: `Are we locking on something else that would allow this to happen?`

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
jeyrschabu: Resources are fetched lazily by the notification sender + there is the same locking mechanism we have elsewhere

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
aravindmd: Ignore , i saw that you have the tests based on `resourceType` in `NotificationSenderTest`

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationQueue.kt
emjburns: I think it's weird that `NotificationTask` is an internal class and yet it's referenced by two other files. Maybe best to move to its own file?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
emjburns: I was confused reading this at first because I didn't see the `do/while` and I didn't go back and update my comment. What if we just had a `popAll` method on the queue that returned everything?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
emjburns: oof hit me back with the "what's your suggestion" 😆 good one! I'll think about it and try to offer one on my next pass of PR review.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
emjburns: ahh I read so well. I totally see that now :)

jeyrschabu:
I was already in the code so I cut this guy to cover the echo side once this is merged:

https://github.com/spinnaker/echo/pull/671

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/MetricsSupport.kt
```diff
@@ -23,6 +23,7 @@ import com.netflix.spinnaker.swabbie.events.Action
 import com.netflix.spinnaker.swabbie.model.WorkConfiguration
 import java.util.concurrent.atomic.AtomicInteger
 
+// TODO: Jeyrs - refactor me I hate this!!!!
```
emjburns: 😆

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/discovery/DiscoveryActivated.kt
```diff
@@ -4,15 +4,25 @@ import com.netflix.appinfo.InstanceInfo
...
+import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * A component that starts doing something when the instance is up in discovery
  * and stops doing that thing when it goes down.
  */
-interface DiscoveryActivated : ApplicationListener<RemoteStatusChangedEvent> {
+open class DiscoveryActivated : ApplicationListener<RemoteStatusChangedEvent> {
+  private val up = AtomicBoolean()
+  open fun onDiscoveryUpCallback(event: RemoteStatusChangedEvent) {
```
emjburns: I think these should remain callbacks with no body, and the up/down logic should go in the `onApplicationEvent` function. That way, you can override these to provide specific logic without messing up the up/down indication or having to call `super`.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
```diff
@@ -119,6 +120,22 @@ interface Identifiable : Named {
   val resourceType: String
   val cloudProvider: String
   val grouping: Grouping?
+
+  /**
+   * Returns the opt-out url for this resource
+   * This is used for notifications
+   */
+  fun optOutUrl(workConfiguration: WorkConfiguration): String {
```
emjburns: Does this have any impact on json serialization/deserialization? aka will making changes here lead to problems reading from the current database when we roll it out?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
```diff
@@ -0,0 +1,220 @@
...
+   * The frequency of this scheduled action is cron based and configurable
+   * Defaults to every day at 9am.
+   */
+  @Scheduled(cron = "\${swabbie.notification.cron.schedule:0 0 9 1/1 * ?}")
+  fun sendNotifications() {
+    if (!isUp()) {
+      return
+    }
+
+    lockingService.acquireLock(lockOptions) {
```
emjburns: This is so much clearer to me. Thanks!

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
```diff
@@ -48,37 +47,28 @@ class WorkQueueManager(
...
-        ensureLoadedCaches()
-        queue.refillOnEmpty()
-      }
-    }
-
-  override val onDiscoveryDownCallback: (event: RemoteStatusChangedEvent) -> Unit
-    get() = {
-      up.set(false)
+  override fun onDiscoveryUpCallback(event: RemoteStatusChangedEvent) {
+    if (isEnabled()) {
```
emjburns: as the code is now, this `override` makes it so `DiscoveryActivated` never does the up/down business. see my earlier comment in that class.

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/ResourceTypeHandlerTest.kt
```diff
@@ -652,84 +653,76 @@ object ResourceTypeHandlerTest {
...
+    ) doReturn listOf(markedResource)
+
+    defaultHandler.setCandidates(mutableListOf(defaultResource))
+    defaultHandler.setRules(defaultRules)
+
     defaultHandler.notify(configuration)
-    verify(notifier, times(0)).notify(any(), any(), any())
-    verify(resourceRepository, times(1)).upsert(updatedMarkedResource)
+
+    with(markedResource.notificationInfo) {
```
emjburns: You can use an expect block in strikt, which i think will make this better. I don't want to have to assert that `this` is not null, reads kind of weird in the test.

You might just be able to do:
```kotlin
val notificationInfo = markedResource.notificationInfo
expect {
  that(notificationInfo).isNotNull()
  ...
  that(notificationInfo.NotificationStamp).isNull()
}
```

swabbie-redis/src/test/kotlin/com/netflix/spinnaker/swabbie/redis/RedisNotificationQueueTest.kt
```diff
@@ -0,0 +1,101 @@
...
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+import redis.clients.jedis.JedisPool
+import strikt.api.expectThat
+import strikt.assertions.isEqualTo
+import strikt.assertions.isFalse
+import strikt.assertions.isTrue
+import strikt.assertions.size
+
```
emjburns: Something that Fletch does that I really like is he sets up a base test class (for example, [ResourceRepositoryTests](https://github.com/spinnaker/keel/blob/master/keel-core-test/src/main/kotlin/com/netflix/spinnaker/keel/persistence/ResourceRepositoryTests.kt)) and then each repo type (in memory, sql, redis, etc) extends that test to provide whatever initialization logic they need [InMemory](https://github.com/spinnaker/keel/blob/master/keel-core/src/test/kotlin/com/netflix/spinnaker/keel/persistence/memory/InMemoryResourceRepositoryTest.kt), [sql](https://github.com/spinnaker/keel/blob/master/keel-sql/src/test/kotlin/com/netflix/spinnaker/keel/sql/SqlResourceRepositoryTests.kt)).

This makes it so that you're _sure_ the in memory and real implementations work the same. Something to keep in mind for the future!

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/discovery/DiscoveryActivated.kt
jeyrschabu: Yes, the up/down can go in the onApplication body but I prefer these as overridable functions rather than properties though. The end result would be the same but reads better as functions.

swabbie-redis/src/test/kotlin/com/netflix/spinnaker/swabbie/redis/RedisNotificationQueueTest.kt
jeyrschabu: Got plans to write a proper `tck` at some point.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
jeyrschabu: I dont believe so. Will test that out.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/discovery/DiscoveryActivated.kt
jeyrschabu: grr can't keep that without changing a bunch other stuff. Will keep them as getters

jeyrschabu:
Ok will update this with the feedback re:discoveryActivated and mergeroni. Likely to tune things out depending how testing goes.

PR #259, opened 2019-10-08
feat(asg): Update ASG rule to account for number of days resource is disabled

aravindmd:
Updates to the ASG violation rule to taken into account the number of days the resource has been disabled


swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
```diff
@@ -16,31 +16,65 @@
...
+
+  fun isOutOfLoadBalancer(): Boolean {
+    suspendedProcesses?.find {
+      it.processName == "AddToLoadBalancer"
+    }?.let {
+      return true
+    }
+
+    return false
+  }
```
robzienert: Or...

```kotlin
fun isOutOfLoadBalancer(): Boolean = suspendedProcesses
  ?.anyMatch { it.processName == "AddToLoadBalancer" }
```

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
```diff
@@ -16,31 +16,65 @@
...
+
+    return null
+  }
+
+  fun disabledTime(): LocalDateTime? {
+    if (!isOutOfLoadBalancer()) {
+      return null
+    }
+
+    val timeSinceDisabled = REGEX.find(suspensionReason()!!)!!.groups.get(1)!!.value
```
robzienert: Code smell if you're `!!` everywhere.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
```diff
@@ -16,31 +16,65 @@
...
   val loadBalancerNames: List<String>?,
+  var suspendedProcesses: List<SuspendedProcess>? = null,
   private val createdTime: Long,
   override val resourceId: String = autoScalingGroupName,
   override val resourceType: String = SERVER_GROUP,
   override val cloudProvider: String = AWS,
   override val name: String = autoScalingGroupName,
   private val creationDate: String? = LocalDateTime.ofInstant(Instant.ofEpochMilli(createdTime), ZoneId.systemDefault()).toString()
 ) : AmazonResource(creationDate) {
+  private val REGEX = "(\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}).*".toRegex()
```
jeyrschabu: Let's name this something specific. Like suspensionReasonDateRegex

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/Rules.kt
```diff
@@ -48,23 +60,28 @@ class ZeroInstanceDisabledServerGroupRule : Rule<AmazonAutoScalingGroup> {
...
+
+  @Value("\${swabbie.resource.disabled.days.count:30}")
+  private val disabledDurationInDays: Long = 30
+
   override fun apply(resource: AmazonAutoScalingGroup): Result {
     if (resource.details[IS_DISABLED] == null || resource.details[IS_DISABLED] == false) {
       return Result(null)
     }
-
+    val disabledTime = resource.disabledTime() ?: return Result(null)
```
jeyrschabu: We should get the lastUpdatedTimestamp on the instance info instead.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
aravindmd: refactored based on some of your comments , let me know

aravindmd:
> Good start!
> 
> Left a comment on using the discovery timestamp instead of the lb disabled time.

Updated the code with some minor refactor , take a look and let me know , there are still a couple of methods which i would like to revisit later

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/Rules.kt
```diff
@@ -17,27 +17,43 @@
...
-    if (resource.details[HAS_INSTANCES] == null || resource.details[HAS_INSTANCES] == false) {
-      return Result(
-        Summary(
-          description = "Server Group ${resource.autoScalingGroupName} is disabled. No active instances.",
-          ruleName = javaClass.simpleName
+    if (!resource.hasInstances()) {
+      val disabledTime = resource.disabledTime() ?: return Result(null)
+
+      // time elapsed since the resource was disabled time is greater than disabledDurationInDays
+      if (disabledTime.isBefore(LocalDateTime.now().minusDays(disabledDurationInDays))) {
```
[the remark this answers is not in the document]
aravindmd: updated to use clock

swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/ZeroInstanceInDiscoveryDisabledServerGroupRuleTest.kt
```diff
@@ -59,19 +65,27 @@ object ZeroInstanceInDiscoveryDisabledServerGroupRuleTest {
     whenever(instanceInfo.status) doReturn
       InstanceInfo.InstanceStatus.OUT_OF_SERVICE
 
+    whenever(instanceInfo.lastUpdatedTimestamp) doReturn "147049205495".toLong()
```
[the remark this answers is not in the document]
aravindmd: fixed this , should be readable now

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
```diff
@@ -16,31 +16,69 @@
...
+
   override fun equals(other: Any?): Boolean {
     return super.equals(other)
   }
 
   override fun hashCode(): Int {
     return super.hashCode()
   }
+
+  fun isOutOfLoadBalancer(): Boolean {
```
emjburns: you call this at least twice as `!isOutOfLoadBalancer()`. Do you think it's better to switch it to the opposite - `isInLoadBalancer()` (and flip the condition)

swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroupHandlerTest.kt
```diff
@@ -82,7 +82,7 @@ object AmazonAutoScalingGroupHandlerTest {
   private val resourceStateRepository = mock<ResourceStateRepository>()
   private val taskTrackingRepository = mock<TaskTrackingRepository>()
   private val resourceOwnerResolver = mock<ResourceOwnerResolver<AmazonAutoScalingGroup>>()
-  private val clock = Clock.fixed(Instant.parse("2018-05-24T12:34:56Z"), ZoneOffset.UTC)
+  private val clock = Clock.systemUTC()
```
emjburns: does this switch the clock to a non-fixed clock? If so, how come?

swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/ZeroInstanceDisabledServerGroupRuleTest.kt
```diff
@@ -16,11 +16,17 @@
...
+import com.amazonaws.services.autoscaling.model.SuspendedProcess
 import org.junit.jupiter.api.Assertions
 import org.junit.jupiter.api.Test
+import java.time.Clock
 import java.time.Instant
+import java.time.temporal.ChronoUnit
 
 object ZeroInstanceDisabledServerGroupRuleTest {
+
+  private val clock = Clock.systemUTC()
```
emjburns: Clocks in tests should use our [`MutableClock`](https://github.com/spinnaker/kork/blob/master/kork-test/src/main/java/com/netflix/spinnaker/kork/test/time/MutableClock.java) so that we can set and increment them. We don't want to be relying on a system clock.

emjburns: That goes for every test clock in this PR.
If you don't need to increment the clock at all, you can use the `Clock.fixed(...)` method.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
```diff
@@ -16,31 +16,69 @@
...
+    return (details[HAS_INSTANCES] != null && details[HAS_INSTANCES] == true)
+  }
+
+  // TODO :aravindd refactor this method
+  fun disabledTime(): LocalDateTime? {
+    if (!isOutOfLoadBalancer()) {
+      return null
+    }
+
+    val suspensionReason = getLoadBalancerSuspensionReason()!!
```
emjburns: Are you 100% certain this will never be null? I'm not. Better to handle the null case by throwing an exception or logging rather than asserting this will never be null with `!!`

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
```diff
@@ -16,31 +16,69 @@
...
+  fun isDisabled(): Boolean {
+    return (details[IS_DISABLED] != null && details[IS_DISABLED] == true)
+  }
+
+  fun hasInstances(): Boolean {
+    return (details[HAS_INSTANCES] != null && details[HAS_INSTANCES] == true)
+  }
+
+  // TODO :aravindd refactor this method
+  fun disabledTime(): LocalDateTime? {
```
emjburns: I think it's worth noting (for future readers) that the disabled time comes from Amazon and is trustworthy. It would probably be good to link to the amazon docs as well for the relevant information.

swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/ZeroInstanceDisabledServerGroupRuleTest.kt
jeyrschabu: I think a system clock is fine if you don’t need to mutate the clock to test a piece of functionality.
But this requires your test to use this clock everywhere to make relative calculations of time, because the class under test relies on that same clock.

However looking more in the test, this clock isn’t used everywhere: this makes test brittle if you are using LocalDateTime.now() and not use the clock to match the class under test.

aravindmd:
Updated the tests to use the system `clock` and pass them in the required places , did not use `Mutable` clock since the tests are not modifying the time cc @emjburns @jeyrschabu

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
aravindmd: added a note , did not find the any relevant AWS doc with details . If you know of a doc which explains this better let me know i will link it here

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
aravindmd: I think it's easier to read this way , i will refactor if we feel flipping the condition is better.

swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/ZeroInstanceDisabledServerGroupRuleTest.kt
emjburns: I think any use of the actual clock is a recipe for tests to become brittle, fail at certain times, hard to diagnose. I'd much rather use a fixed clock. It's just as easy and much safer.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
emjburns: I personally think it's harder to read when you're always negating something. It's a sign to me that your function is wrong if you always want the opposite of the result.

aravindmd:
Discussed with @emjburns about the clock usage , for this set of tests its ok to use fixed clock , going to look at Mutable clock during the next refactor.

PR #270, opened 2019-10-23
Endpoint to Trigger Notifications for all resource in the notification queue

aravindmd:
- Endpoint to Trigger Notifications for all resource in the notification queue
- Added a endpoint on the testing controller


aravindmd:
cc @jeyrschabu @emjburns

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/OnDemandNotifyResourceData.kt
```diff
@@ -0,0 +1,32 @@
+package com.netflix.spinnaker.swabbie.model
+
+import com.fasterxml.jackson.annotation.JsonTypeName
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+
+data class OnDemandNotifyResourceData(
```
jeyrschabu: Let's keep one of these only. We don't need to create a new one.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/OnDemandNotifyResourceData.kt
```diff
@@ -0,0 +1,32 @@
...
+  val resourceId: String,
+  var projectedDeletionStamp: Long,
+  var markTs: Long?,
+  var resourceOwner: String = "swabbie@spinnaker.io",
+  val account: String,
+  val location: String
+)
+
+@JsonTypeName("testResource")
+data class NotifyResource(
```
jeyrschabu: Use the testResource for this

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/notifications/NotificationSender.kt
```diff
@@ -220,7 +220,7 @@ class NotificationSender(
       .monitorValue(size())
   }
 
-  private data class NotificationResourceData(
```
jeyrschabu: nit: I'd prefer not leaking this class out. Let's keep it private and build out what the notificationSender requires from the caller code. (Not needed for now but should refactor after testing)

swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/TestingController.kt
```diff
@@ -86,6 +92,37 @@ class TestingController(
...
+    @PathVariable namespace: String,
+    @RequestBody resource: OnDemandNotifyResourceData
+  ) {
+    val workConfiguration = findWorkConfiguration(SwabbieNamespace.namespaceParser(namespace))
+    val markedResource = createMarkedResource(workConfiguration,
+      resource.resourceId,
+      resource.resourceOwner,
+      resource.projectedDeletionStamp,
+      resource.markTs)
+    val notificationResourceData = NotificationSender.NotificationResourceData(
```
jeyrschabu: see comment on keeping this class private. (for later refactor)

swabbie-web/src/main/kotlin/com/netflix/spinnaker/swabbie/controllers/TestingController.kt
```diff
@@ -94,4 +131,25 @@ class TestingController(
...
+
+  private fun createMarkedResource(
+    workConfiguration: WorkConfiguration,
+    id: String,
+    owner: String,
+    projectedDeletionStamp: Long,
+    markTs: Long?
+  ): MarkedResource {
+    return MarkedResource(
+      resource = NotifyResource(resourceId = id,
```
jeyrschabu: Use testResource

aravindmd:
removed `OnDemandNotifyResourceData` using 'OnDemandMarkData' instead . Also updated to use `testResource`

aravindmd:
> Going to merge this as-is @aravindmd. I'd like a followup to address the remaining items

I did not want to change this interface for other endpoints , but since they are mostly for testing i guess i can make it required. I'll update that with the next change.

PR #273, opened 2019-10-24
chore(metrics): Adding some db failure metrics

jeyrschabu:
- tracking failures to read data from the database


swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisResourceTrackingRepository.kt
```diff
@@ -197,12 +190,7 @@ class RedisResourceTrackingRepository(
     try {
       markedResource = objectMapper.readValue(resource)
     } catch (e: Exception) {
-      registry.counter(
-        redisErrors.withTags(
-          "exception", e.javaClass.simpleName,
-          "action", "readMarkedResource"
-        )
-      ).increment()
+      registry.counter(registry.createId("redis.resourceTracking.errors")).increment()
```
emjburns: Worth pulling the creation of this ID up to the top of the class like we do elsewhere, so we don't "create" it twice?

jeyrschabu: Late night PRs... saw that this morning.

Issue #274, opened 2019-10-24
Fix swabbie email template termination date


jeyrschabu:
<img width="878" alt="Screen Shot 2019-10-24 at 1 10 53 PM" src="https://user-images.githubusercontent.com/1572759/67522109-f937fc80-f660-11e9-9059-9a0d42e31495.png">

<img width="511" alt="Screen Shot 2019-10-24 at 1 12 41 PM" src="https://user-images.githubusercontent.com/1572759/67522116-ff2ddd80-f660-11e9-9ee6-437a30f3212b.png">

Echo template: ```<td align="left" style="padding: 4px 0; font-family: Helvetica, Arial, sans-serif; font-size: 12px;">
                    ${resourceData.resource.projectedDeletionStamp?number_to_date?string("EEE, d MMM yyyy")}
                  </td>```

Done When:
- Swabbie formats the deletion date in human readable format
- Update echo to read the value without any treatment to the format

PR #276, opened 2019-10-25
chore(notifications): Ensure deletion date is passed to echo

jeyrschabu:
- passing a human readable date for echo swabbie email template


jeyrschabu:
ref: https://github.com/spinnaker/echo/pull/687

jeyrschabu:
Addresses https://github.com/spinnaker/swabbie/issues/274

PR #278, opened 2019-10-27
fix(core): Removing bulk unmarking of marked resources

jeyrschabu:
- remove unnecessary bulk unmarking
- fix test with previous false positive


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -327,13 +288,6 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
         }
       }
 
-      try {
```
jeyrschabu: The "Previously marked resources that are no longer violating rules" needing to be unmarked is already covered here https://github.com/spinnaker/swabbie/blob/master/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt#L291

PR #279, opened 2019-10-28
Namespace case fix while reading from redis

aravindmd:
- convert `namespace` to lower case while reading from redis , Eg for resource like `server group` the namespace is stored as `aws:test:us-east-1:servergroup` in redis and Gate requests have resource type as `serverGroup`


aravindmd:
@jeyrschabu

PR #282, opened 2019-11-04
fix(exclusions): Changing Exclusion attribute value type

jeyrschabu:
- Change Exclusion.Attribute.value from `List<String>` to `List<Any>`

### DIFF
--- a/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/exclusions/AmazonTagExclusionPolicy.kt
+++ b/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/exclusions/AmazonTagExclusionPolicy.kt
@@ -47,8 +47,8 @@ class AmazonTagExclusionPolicy(
     val configuredKeysAndTargetValues = keysAndValues(exclusions, ExclusionType.Tag)
     tags.forEach { tag ->
       val target = configuredKeysAndTargetValues[tag.key] ?: emptyList()
-      if (target.contains(tag.value as String)) {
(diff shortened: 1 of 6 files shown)


swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/exclusions/AmazonTagExclusionPolicy.kt
```diff
@@ -47,8 +47,8 @@ class AmazonTagExclusionPolicy(
     val configuredKeysAndTargetValues = keysAndValues(exclusions, ExclusionType.Tag)
     tags.forEach { tag ->
       val target = configuredKeysAndTargetValues[tag.key] ?: emptyList()
-      if (target.contains(tag.value as String)) {
```
jeyrschabu: This cast would fail if the tag value is not a string.

emjburns: Do you want to handle things differently if it's not a string? Ignore or throw an error? I guess, what is the object that is failing to be cast to a string, but calling `.toString()` on it gives a reasonable behavior?

jeyrschabu: I definitely wouldn't want to ignore because it would mean we probably are letting resources in that should be excluded. In case of a failure, the exception would bubble up, logged and the action would be aborted.

I prefer making the exclusion mechanism explicit rather than stringifying the value.

PR #284, opened 2019-11-04
refactor(core): Move queue metrics to queues

jeyrschabu:
- moved track method for queues to respective impl


jeyrschabu:
cc: @aravindmd

swabbie-redis/src/main/kotlin/com/netflix/spinnaker/swabbie/redis/RedisWorkQueue.kt
```diff
@@ -40,6 +43,7 @@ class RedisWorkQueue(
 
   init {
     log.info("Using ${javaClass.simpleName}")
+    this.track()
```
aravindmd: What about this , shouldn't we get rid of this now ?
https://github.com/spinnaker/swabbie/blob/master/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt#L118

PR #290, opened 2019-11-06
fix(events): Dont resend additional events

jeyrschabu:
- remove unnecessary eventing in orca task agent


swabbie-orca/src/main/kotlin/com/netflix/spinnaker/swabbie/orca/OrcaTaskMonitoringAgent.kt
```diff
@@ -174,23 +170,6 @@ class OrcaTaskMonitoringAgent(
     taskTrackingRepository.cleanUpFinishedTasks(daysToKeepTasks)
   }
 
-  private fun publishEvent(taskInfo: TaskCompleteEventInfo) {
-    when (taskInfo.action) {
-      Action.DELETE -> {
-        taskInfo.markedResources
-          .forEach { markedResource ->
-            applicationEventPublisher.publishEvent(DeleteResourceEvent(markedResource, taskInfo.workConfiguration))
```
jeyrschabu: This is handled here https://github.com/spinnaker/swabbie/blob/master/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceStateManager.kt#L87

jeyrschabu:
@aravindmd

PR #292, opened 2019-11-06
fix(core): Fixing optout flow

jeyrschabu:
- moved cloud provider opt out actions to handler
- update handler interface optout function
- fix abstraction for optout flow


jeyrschabu:
@aravindmd per our discussion this PR fixes issues we are seeing with the optout flow.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -149,39 +149,43 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
...
-    // this will trigger the actual opt out action
-    applicationEventPublisher.publishEvent(OptOutResourceEvent(newMarkedResource, workConfiguration))
+    log.info("Opting out resource ${markedResource.resourceId} in namespace ${workConfiguration.namespace}")
+    ensureResourceUnmarked(markedResource, workConfiguration, "Opted Out")
+    applicationEventPublisher.publishEvent(OptOutResourceEvent(markedResource, workConfiguration))
+    return optOutState.copy(
+      currentStatus = status,
+      statuses = setOf(optOutState.statuses + status).flatten().toMutableList()
+    ).apply {
+      resourceStateRepository.upsert(this)
```
aravindmd: what about `resourceRepository.upsert()` ? will this take care of resource upsert ?

jeyrschabu: Why would we do that while we are ensuring this resource is opted out & umarked?

aravindmd: ok so we don't need to update the resource state , i ask this because the previous optout flow had this call?

jeyrschabu: That was definitely a bug. resourceRepository.upsert would end up marking the resource. ResourceStateRepository.upsert does update the status to opted out which is what we need

PR #293, opened 2019-11-06
fix(serverGroup): Fixing asg rules and handler

jeyrschabu:
- remove misleading classifiers for disabled. use fields instead
- fix test

### DIFF
--- a/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
+++ b/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
@@ -79,14 +79,6 @@ data class AmazonAutoScalingGroup(
     return getAddToLoadBalancerProcess()?.suspensionReason
   }
 
-  fun isDisabled(): Boolean {
-    return (details[IS_DISABLED] != null && details[IS_DISABLED] == true)
-  }
-
-  fun hasInstances(): Boolean {
-    return (details[HAS_INSTANCES] != null && details[HAS_INSTANCES] == true)
-  }
-
   // TODO :aravindd refactor this method
   // Disabled time is here is provided by AWS
   fun disabledTime(): LocalDateTime? {
--- a/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroupHandler.kt
+++ b/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroupHandler.kt
@@ -115,10 +115,9 @@ class AmazonAutoScalingGroupHandler(
     }
   }
 
-  override fun handles(workConfiguration: WorkConfiguration):
-    Boolean = workConfiguration.resourceType == SERVER_GROUP &&
-    workConfiguration.cloudProvider == AWS &&
-    !rules.isEmpty()
+  override fun handles(workConfiguration: WorkConfiguration): Boolean {
+    return workConfiguration.resourceType == SERVER_GROUP && workConfiguration.cloudProvider == AWS && rules.isNotEmpty()
+  }
 
   override fun getCandidates(workConfiguration: WorkConfiguration): List<AmazonAutoScalingGroup>? {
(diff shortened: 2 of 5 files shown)


swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
```diff
@@ -79,14 +79,6 @@ data class AmazonAutoScalingGroup(
     return getAddToLoadBalancerProcess()?.suspensionReason
   }
 
-  fun isDisabled(): Boolean {
```
jeyrschabu: @aravindmd this creates a circular dependency.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
```diff
@@ -79,14 +79,6 @@ data class AmazonAutoScalingGroup(
...
-    return (details[IS_DISABLED] != null && details[IS_DISABLED] == true)
-  }
-
-  fun hasInstances(): Boolean {
```
jeyrschabu: See above comment

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroupHandler.kt
```diff
@@ -152,20 +151,7 @@ class AmazonAutoScalingGroupHandler(
     if (serverGroups == null || serverGroups.isEmpty()) {
       return
     }
-    val elapsedTimeMillis = measureTimeMillis {
```
jeyrschabu: No need to do this here since. We have all references already present. Load balancer and instances

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/Rules.kt
```diff
@@ -72,7 +72,7 @@ class ZeroInstanceInDiscoveryDisabledServerGroupRule(
   private val disabledDurationInDays: Long = 30
 
   override fun apply(resource: AmazonAutoScalingGroup): Result {
-    if (!resource.isDisabled()) {
+    if (resource.isInLoadBalancer()) {
```
jeyrschabu: more explicit

swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/ZeroInstanceDisabledServerGroupRuleTest.kt
```diff
@@ -35,12 +35,9 @@ object ZeroInstanceDisabledServerGroupRuleTest {
       instances = listOf(
         mapOf("instanceId" to "i-01234")
       ),
-      loadBalancerNames = listOf(),
+      loadBalancerNames = listOf("lb1"),
       createdTime = clock.millis()
-
-    ).apply {
-      set(IS_DISABLED, false)
```
jeyrschabu: @aravindmd better to check actual fields on the server groups.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/AmazonAutoScalingGroup.kt
aravindmd: Can you explain this bit more ?

jeyrschabu: `IS_DISABLED` is not a field on the amazon autoscaling group. It's something that the handler was responsible to set.

aravindmd: Thanks , that makes sense now.

PR #295, opened 2019-11-07
fix(exclusion): Do not excluded expired resources

jeyrschabu:
- fixed to not expire if temporal tag has value "never"
- Updated to not exclude an expired resource


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
```diff
@@ -94,18 +94,29 @@ abstract class Resource : Excludable, Timestamped, HasDetails() {
   fun getTagValue(key: String): Any? = tags()?.find { it.key == key }?.value
 
   fun expired(clock: Clock): Boolean {
-    return tags()?.any { expired(it, clock) } ?: false
+    tags()?.let { tags ->
+      if (tags.any { it.isTemporal() && it.value == "never" }) {
```
jeyrschabu: @aravindmd temporalTag.value =`never` takes precedence.

aravindmd: :+1: makes sense

PR #297, opened 2019-11-08
feat(*): Initial CODEOWNERS

ajordens:


ajordens:
Seeded off of the previous `OWNERS.md` (not sure if @emjburns wants to be tagged or not?).

PR #302, opened 2019-11-14
fix(core): Ensure resource is removed from db when deleted

jeyrschabu:
- updated to ensure the deleted resource is removed from the db

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
@@ -500,10 +500,14 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
     val maxItemsToProcess = min(confirmedResourcesToDelete.size, workConfiguration.getMaxItemsProcessedPerCycle(dynamicConfigService))
     confirmedResourcesToDelete.subList(0, maxItemsToProcess).let {
       partitionList(it, workConfiguration).forEach { partition ->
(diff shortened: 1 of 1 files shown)


jeyrschabu:
Caught this while adding a tck for handlers: https://github.com/spinnaker/swabbie/pull/303

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -500,10 +500,14 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
...
     confirmedResourcesToDelete.subList(0, maxItemsToProcess).let {
       partitionList(it, workConfiguration).forEach { partition ->
-        if (!partition.isEmpty()) {
+        if (partition.isNotEmpty()) {
           try {
             deleteResources(partition, workConfiguration)
-            partition.forEach { it -> applicationEventPublisher.publishEvent(DeleteResourceEvent(it, workConfiguration)) }
+            partition.forEach { markedResource ->
+              applicationEventPublisher.publishEvent(DeleteResourceEvent(markedResource, workConfiguration))
+              resourceRepository.remove(markedResource)
```
emjburns: Should we remove from the resource state repository as well?

jeyrschabu: Resource states should be pruned at a later time, either by a scheduled action (currently) or expiring keys. I would like to keep an audit trail for some short period of time before deleting.

PR #306, opened 2019-11-24
feat(core): Allow to define enabled rules in configuration

jeyrschabu:
- Updated to eval resource against config specified rules
- falls back on default behavior if not specified in config
- BONUS: Basic support for AND & OR operators to allow rules composition

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
@@ -162,6 +162,32 @@ class ResourceTypeConfiguration {
 
     // generic type for flexibility
     var rules: List<Any> = emptyList()
+
+    override fun equals(other: Any?): Boolean {
+      if (other is RuleConfiguration) {
+        if (operator != other.operator) {
+          return false
+        }
+
+        if (rules != other.rules) {
+          return false
+        }
+
+        return true
+      }
+
+      return false
+    }
+
+    override fun hashCode(): Int {
+      var result = operator.hashCode()
+      result = 31 * result + rules.hashCode()
+      return result
+    }
+
+    override fun toString(): String {
+      return "RuleConfiguration(operator=$operator, rules=$rules)"
+    }
   }
 }
 
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
@@ -133,6 +133,36 @@ class ResourceTypeConfiguration {
   var entityTaggingEnabled: Boolean = false
   var maxAge: Int = 14
   var notification: NotificationConfiguration = EmptyNotificationConfiguration()
+  var enabledRules: List<RuleConfiguration> = listOf()
+
+  /**
+   * Allows to specify enabled rules. When defined, all other rules are ignored at runtime.
+   * See [com.netflix.spinnaker.swabbie.AbstractResourceTypeHandler.getViolations]
+   *------------- config snippet for a resource type---------------
+          enabledRules:
+          - operator: OR
+            rules:
+              - ZeroInstanceInDiscoveryDisabledServerGroupRule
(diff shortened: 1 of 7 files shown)


jeyrschabu:
In an effort to improve extensibility and in preparation for plugins support, this PR adds some much needed flexibility to how rules are composed and applied to resources.

This means generic rules can be interchangeably applied to different resource types if needed, thus reducing the need to create rigid custom rules that may not be applicable for everyone in OSS.
Smaller rules can be created and composed using an operator such as `AND` and `OR`:
`ZeroInstanceRule`
`NoApplicationRule`
`NoServerGroupRule`
...

Additionally, this gives yet another control knob into swabbie actions.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -552,6 +553,54 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
...
 
+  /**
+   * Evaluates a resource against rules and returns violation summaries for this resource
+   * @see [com.netflix.spinnaker.config.ResourceTypeConfiguration.RuleConfiguration]
+   * @see [WorkConfiguration.enabledRules]
+   */
+  private fun T.getViolations(workConfiguration: WorkConfiguration): List<Summary> {
+    if (workConfiguration.enabledRules.isNullOrEmpty()) {
+      // Evaluate the resource against all rules and get violations
+      return getViolations()
```
aravindmd: do we want to evaluate all the all rules in this case vs not run any of them , trying to understand the reason here ?

swabbie-core/src/main/kotlin/com/netflix/spinnaker/config/SwabbieProperties.kt
```diff
@@ -133,6 +133,36 @@ class ResourceTypeConfiguration {
   var entityTaggingEnabled: Boolean = false
   var maxAge: Int = 14
   var notification: NotificationConfiguration = EmptyNotificationConfiguration()
+  var enabledRules: List<RuleConfiguration> = listOf()
+
+  /**
```
aravindmd: :+1:

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
jeyrschabu: If there are no enabled rules it will fallback to the usual behavior. I added unit tests to illustrate that.

swabbie-core/src/test/kotlin/com/netflix/spinnaker/swabbie/ResourceTypeHandlerTest.kt
```diff
@@ -155,6 +160,103 @@ object ResourceTypeHandlerTest {
...
+    expectThat(
+      defaultHandler.getViolations(resource, workConfiguration.copy(enabledRules = emptyList()))
+    ).isNotEmpty()
+      .get {
+        this.map { it.ruleName }
+      }.isEqualTo(
+        listOf(testRule1.name())
+      )
+
+    // andRuleConfig
```
aravindmd: test with `andRule` resulting in violation would be good

jeyrschabu: Good call sir! Adding this test found a line I lost in one of the commits.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -586,10 +586,12 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
           }
 
           if (allApplied) {
-            resolveRules(rules, ruleConfig)
-              .mapNotNull {
-                it.apply(resource).summary
-              }
+            violationSummaries.addAll(
```
jeyrschabu: Lost that line previously. Test caught it!

PR #315, opened 2019-12-13
feat(rules): Adding some rules

jeyrschabu:
- added a few asg rules
- added an age rule
- updated rules api

### DIFF
--- a/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/Rules.kt
+++ b/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/Rules.kt
@@ -156,6 +156,8 @@ class DisabledLoadBalancerRule(
     }
 
     val disabledTime = resource.disabledTime()!!
+
+    // Parameter moreThanDays is optional. This rule would enforce moreThanDays if present otherwise it just asserts if the server group is disabled
     val moreThanDays = ruleDefinition?.parameters?.get("moreThanDays") as? Int
     if (moreThanDays != null && disabledTime.isBefore(LocalDateTime.now(clock).minusDays(moreThanDays.toLong()))) {
       return Result(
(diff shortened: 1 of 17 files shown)


swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/Rules.kt
```diff
@@ -109,3 +111,122 @@ class ZeroInstanceInDiscoveryDisabledServerGroupRule(
...
+
+    val disabledTime = resource.disabledTime()!!
+    val moreThanDays = ruleDefinition?.parameters?.get("moreThanDays") as? Int
+    if (moreThanDays != null && disabledTime.isBefore(LocalDateTime.now(clock).minusDays(moreThanDays.toLong()))) {
+      return Result(
+        Summary(description = "Server Group ${resource.resourceId} has been out of balancer longer than $moreThanDays days.", ruleName = name())
+      )
+    }
+
+    return Result(
```
aravindmd: not clear why is this needed ?

jeyrschabu: `moreThanDays` is optional in this rule. This rule would enforce it if present otherwise it just asserts if the server group is disabled

jeyrschabu: Will add that as a comment.

aravindmd: gotcha !

swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/autoscalinggroups/DisabledLoadBalancerRuleTest.kt
```diff
@@ -0,0 +1,70 @@
...
+      }
+
+    expectThat(rule.apply(asg, ruleDefinition).summary).isNull()
+
+    ruleDefinition = RuleDefinition()
+      .apply {
+        name = rule.name()
+        parameters = mapOf("moreThanDays" to 1)
+      }
+
```
aravindmd: maybe worth adding the case for optional `moreThanDays` constraint here , which will validate just the serverGroup being disabled

jeyrschabu: Reasonable. Will queue that up in another in-flight PR.

PR #317, opened 2019-12-13
feat(rules): Attribute Rule

jeyrschabu:
- added an attribute rule that would apply if specified parameters match
the resource attributes

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/exclusions/ExclusionPolicy.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/exclusions/ExclusionPolicy.kt
@@ -22,8 +22,6 @@ import com.netflix.spinnaker.swabbie.model.Identifiable
 import com.netflix.spinnaker.swabbie.model.WorkConfiguration
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
-import java.util.NoSuchElementException
-import kotlin.reflect.full.memberProperties
 
 interface ExclusionPolicy {
   val log: Logger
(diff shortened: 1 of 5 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/exclusions/ExclusionPolicy.kt
```diff
@@ -69,45 +67,14 @@ interface ExclusionPolicy {
...
-      findProperty(excludable, key, kv.getValue(key))?.let {
+      excludable.findMatchingAttribute(key, kv.getValue(key))?.let {
         return patternMatchMessage(key, setOf(it))
       }
     }
 
     return null
   }
 
-  fun findProperty(excludable: Excludable, key: String, values: List<Any>): String? {
```
jeyrschabu: Moved to resource level for reuse

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
```diff
@@ -171,6 +173,50 @@ interface Identifiable : Named {
...
+      throw e
+    }
+  }
+
+  private fun <R : Any?> readPropery(propertyName: String): R {
+    @Suppress("UNCHECKED_CAST")
+    return javaClass.kotlin.memberProperties.first { it.name == propertyName }.get(this) as R
+  }
+
+  fun String.patternMatched(p: String): Boolean =
```
aravindmd: I think we can add a comment with a sample rule here , i think it would make easier to read

PR #323, opened 2020-01-03
improvement(attributesRule): Handling non strings in attributeRule

jeyrschabu:
- updated attributes matching to handle non strings properties

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
@@ -174,27 +174,23 @@ interface Identifiable : Named {
     return "${workConfiguration.notificationConfiguration.resourceUrl}/$resourceId"
   }
 
-  fun findMatchingAttribute(key: String, values: List<Any>): String? {
-    try {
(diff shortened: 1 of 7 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/model/Resource.kt
```diff
@@ -174,35 +174,31 @@ interface Identifiable : Named {
...
-
-    return null
   }
 
-  private fun propertyMatches(values: List<Any>, fieldValue: String?): Boolean {
-    if (fieldValue == null) {
-      return false
+  private fun Any?.matchesAny(values: List<Any?>): Boolean {
+    if (this == null || this !is String) {
+      return values.contains(this)
```
aravindmd: why not just `return false` here?

jeyrschabu: We wouldn't want that to return false right. How about non strings or nullables?

aravindmd: 👍 makes sense

PR #324, opened 2020-01-10
chore(redis): Return null on null state string

jeyrschabu:
- return null right away if state is null


jeyrschabu:
There are a couple of null strings from way back in n main that needs to be removed from the db. Will do that later but this is good either way.

PR #340, opened 2020-01-30
fix(exclusion): Should be adding the resource type level exclusions

jeyrschabu:
- Fixed a bug where the provider level exclusions were being merged onto
- Should be adding to the resource type level

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/WorkConfigurator.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/WorkConfigurator.kt
@@ -118,7 +118,7 @@ open class WorkConfigurator(
     return all
   }
 
-  private fun mergeExclusions(global: MutableSet<Exclusion>?, local: Set<Exclusion>?): Set<Exclusion> {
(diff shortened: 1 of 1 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/WorkConfigurator.kt
```diff
@@ -128,16 +128,17 @@ open class WorkConfigurator(
     }
 
     // local is additive to global. local can override global
-    merge(local!!, global!!)
```
jeyrschabu: Goddamn. @aravindmd Reason why there were so few resources being marked. The allow list of the other resource type was bleeding into other resource types.

jeyrschabu:
Going to merge this so I can do one more debugging session before swabbie clocks out for the day.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/WorkConfigurator.kt
jeyrschabu: Old bug but now manifesting because the server group exclusions were different than that of images and snapshots.

aravindmd:
Oh Boy, it makes sense now why locally it was marking fine , i must have
been using different resource list

PR #341, opened 2020-01-31
chore(exclusion): Simplify exclusion merging

jeyrschabu:
- simplify merging: initial assumption no longer applies
- no need for fancy merging attributes

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/WorkConfigurator.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/WorkConfigurator.kt
@@ -119,41 +119,18 @@ open class WorkConfigurator(
   }
 
   private fun mergeExclusions(global: Set<Exclusion>?, local: Set<Exclusion>?): Set<Exclusion> {
-    if ((global == null || global.isEmpty()) && (local == null || local.isEmpty())) {
-      return emptySet()
-    } else if ((global == null || global.isEmpty()) && local != null) {
-      return local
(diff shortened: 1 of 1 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/WorkConfigurator.kt
```diff
@@ -119,41 +119,18 @@ open class WorkConfigurator(
...
       exclusionsProviders.forEach { exclusionProvider ->
-        merge(exclusionProvider.get().toSet(), result)
+        result.addAll(exclusionProvider.get().toSet())
       }
     }
 
-    return result.toSet()
-  }
-
-  private fun merge(from: Set<Exclusion>, to: MutableSet<Exclusion>) {
```
jeyrschabu: Part of an early optimization that is no longer needed

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/WorkConfigurator.kt
```diff
@@ -119,41 +119,18 @@ open class WorkConfigurator(
   }
 
   private fun mergeExclusions(global: Set<Exclusion>?, local: Set<Exclusion>?): Set<Exclusion> {
-    if ((global == null || global.isEmpty()) && (local == null || local.isEmpty())) {
```
jeyrschabu: By doing this, we actually never included the runtime exclusions below. Exiting too early.

jeyrschabu:
Ok, mergeroning this to run a couple of tests.

PR #344, opened 2020-02-12
chore(metrics): Cleaning out metrics

jeyrschabu:
- Updated to remove the need for MetricsSupport
- Cleaned up counters


jeyrschabu:
I noticed some of the metrics weren't reporting associated tags and it quickly became convoluted to track why. Noticed a few places that needed cleanup. Furthermore, it would be better to change the handler api to return a result object that would contain stat info that would be logged in one place, preferably the WorkProcessor. That's a change for a different PR because I didnt want to introduce too many changes here.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -435,35 +400,26 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
...
-    val markedResources = mutableListOf<MarkedResource>()
     if (workConfiguration.dryRun) {
       log.info("Deletion not enabled in dryRun for {}. Skipping...", workConfiguration.toLog())
       return
     }
 
     val currentMarkedResourcesToDelete = workConfiguration.getMarkedResourcesToDelete()
     if (currentMarkedResourcesToDelete.isEmpty()) {
-      log.info("Nothing to delete. Configuration: {}", workConfiguration.toLog())
+      log.debug("Nothing to delete. Configuration: {}", workConfiguration.toLog())
```
jeyrschabu: Changing these to debug. Too noisy.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/events/ResourceTrackingManager.kt
```diff
@@ -30,11 +28,11 @@ import org.springframework.stereotype.Component
  * Handles any further change in state after a resource is updated
  * for resources that are changed via an orca task.
  */
+// TODO: (jeyrs) - Review me. Don't think this class is needed since the resource removal happens on site
```
jeyrschabu: Left this untouched in this PR.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkProcessor.kt
```diff
@@ -76,31 +79,35 @@ class WorkProcessor(
...
+      } catch (e: Exception) {
+        log.error("Failed to process: {}", work, e)
+        track(work, success = false)
       }
-    } catch (e: Exception) {
-      log.error("Failed to process: {}", work, e)
-      work.track(success = false)
     }
+
+    registry.timer(workDurationId).record(elapsedTimeMillis, TimeUnit.MILLISECONDS)
```
jeyrschabu: Prefer using the timer here with the advantage that all actions are now accounted for. We used to just measure mark duration.

PR #345, opened 2020-02-14
feat(aws): Adding launch configurations cleanup support

jeyrschabu:
- added launch configs support

### DIFF
--- a/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/launchconfigurations/AmazonLaunchConfigurationHandler.kt
+++ b/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/launchconfigurations/AmazonLaunchConfigurationHandler.kt
@@ -1,0 +1,163 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.swabbie.aws.launchconfigurations
+
+import com.netflix.spectator.api.Registry
+import com.netflix.spinnaker.config.SwabbieProperties
+import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
+import com.netflix.spinnaker.swabbie.AbstractResourceTypeHandler
+import com.netflix.spinnaker.swabbie.ResourceOwnerResolver
+import com.netflix.spinnaker.swabbie.aws.AWS
+import com.netflix.spinnaker.swabbie.aws.Parameters
+import com.netflix.spinnaker.swabbie.exclusions.ResourceExclusionPolicy
+import com.netflix.spinnaker.swabbie.model.AWS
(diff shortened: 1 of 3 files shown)


jeyrschabu:
Delete logic coming next but this can be merged to collect some metrics.

jeyrschabu:
Clouddriver op support: https://github.com/spinnaker/clouddriver/pull/4324

jeyrschabu:
Orca support: https://github.com/spinnaker/orca/pull/3442

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/launchconfigurations/AmazonLaunchConfigurationHandler.kt
```diff
@@ -0,0 +1,198 @@
...
+        description = "Deleting LaunchConfigurations: $launchConfigurationNames"
+      )
+    ).let { taskResponse ->
+      val completeEvent = TaskCompleteEventInfo(
+        action = Action.DELETE,
+        markedResources = markedResources,
+        workConfiguration = workConfiguration,
+        submittedTimeMillis = clock.instant().toEpochMilli()
+      )
+
```
aravindmd: Might be worth adding a log message with the resource / orca task id for debugging ?

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/launchconfigurations/AmazonLaunchConfigurationHandler.kt
```diff
@@ -0,0 +1,198 @@
...
+    val launchConfigurationNames = markedResources
+      .map {
+        it.resourceId
+      }.toSet()
+
+    orcaService.orchestrate(
+      OrchestrationRequest(
+        application = application,
+        job = listOf(
+          OrcaJob(
```
aravindmd: do we need to introduce a wait like how we do for images ? not sure the # of launch configs we will start deleting

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/launchconfigurations/AmazonLaunchConfigurationHandler.kt
```diff
@@ -0,0 +1,198 @@
...
+   * Server Groups
+   */
+  private fun checkReferences(launchConfigurations: List<AmazonLaunchConfiguration>?, params: Parameters) {
+    if (launchConfigurations.isNullOrEmpty()) {
+      return
+    }
+
+    log.debug("checking references for {} launch configs. Parameters: {}", launchConfigurations.size, params)
+    val elapsedTimeMillis = measureTimeMillis {
+      setServerGroupReferences(launchConfigurations, params)
```
aravindmd: can this throw an exception ? do we need to wrap this with a try/catch ?

jeyrschabu: Exceptions would bubble up and get handled in `AbstractResourceTypeHandler`

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/launchconfigurations/AmazonLaunchConfigurationHandler.kt
jeyrschabu: Let's collect some metrics first.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/launchconfigurations/AmazonLaunchConfigurationHandler.kt
jeyrschabu: That's a good idea. I would like to stick that in a common place like so it's called for every handler when a task is submitted. I will do that in a seperate PR.

PR #351, opened 2020-02-26
chore(*): Generify snapshot handler

jeyrschabu:
- updated snapshot handler to not be specific to netflix
* prefer using exclusions or rules in config than hard coded
* use AttributeRule if need to match against the snapshot attributes
- null descriptions should be allowed and should be excluded by policy in config

### DIFF
--- a/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/snapshots/AmazonSnapshotHandler.kt
+++ b/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/snapshots/AmazonSnapshotHandler.kt
@@ -29,7 +29,6 @@ import com.netflix.spinnaker.swabbie.events.Action
 import com.netflix.spinnaker.swabbie.exclusions.ResourceExclusionPolicy
 import com.netflix.spinnaker.swabbie.model.AWS
 import com.netflix.spinnaker.swabbie.model.MarkedResource
-import com.netflix.spinnaker.swabbie.model.NAIVE_EXCLUSION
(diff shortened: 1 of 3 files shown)


swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/snapshots/AmazonSnapshotHandler.kt
```diff
@@ -172,35 +171,9 @@ class AmazonSnapshotHandler(
...
-      } else {
-        snapshot.set(NAIVE_EXCLUSION, true)
+      if (usedResourceRepository.isUsed(SNAPSHOT, snapshot.snapshotId, "aws:${params.region}:${params.account}")) {
+        snapshot.set(IMAGE_EXISTS, true)
       }
     }
   }
-
-  // todo eb: generalize this netflix-specific pattern
-  fun descriptionIsFromAutoBake(description: String): Boolean {
```
aravindmd: Where is this check happening instead , do we not need this for `nflx` case ?

aravindmd: never mind , did the read the description before.

jeyrschabu: This check is now replaced by an Attribute rule matching on the description. Having it hard coded here doesn’t allow non Netflix users to use this handler.

PR #356, opened 2020-02-27
fix(*): Fixing npe in elb handler

jeyrschabu:
- fixed an NPE when checking for server group references


swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/loadbalancers/AmazonLoadBalancerHandler.kt
```diff
@@ -162,29 +161,30 @@ class AmazonLoadBalancerHandler(
...
+      loadBalancers.forEach {
+        it.set(isAttachedToServerGroups, serverGroup.loadBalancerNames?.contains(it.name) ?: false)
       }
     }
 
     return loadBalancers
   }
 
-  private fun List<AmazonElasticLoadBalancer>.addServerGroupReferences(serverGroup: AmazonAutoScalingGroup) = filter {
-    (serverGroup.details["loadBalancerNames"] as List<String>).contains(it.name)
```
jeyrschabu: `loadBalancerNames` is optional and would bomb here if null

PR #376, opened 2020-03-18
chore(*): Allow a placeholder account with the empty account

jeyrschabu:
- updated default values to empty account
- made Resource.getProperty public
- updated work configurator to account for the placeholder account


jeyrschabu:
Merging this to test some -nflx stuff

PR #392, opened 2020-04-02
fix(notifications): Preventing resubmitting notifications

jeyrschabu:
- remove retrySupport use from echo notifier
- only allow one worker to send notifications at a time

### DIFF
--- a/swabbie-echo/src/main/kotlin/com/netflix/spinnaker/swabbie/echo/EchoNotifier.kt
+++ b/swabbie-echo/src/main/kotlin/com/netflix/spinnaker/swabbie/echo/EchoNotifier.kt
@@ -42,10 +42,11 @@ class EchoNotifier(
         when {
           notificationType.equals(NotificationType.EMAIL.name, true) -> {
             return try {
+              log.info("Sending notification to $recipient. context: $notificationContext")
               sendEmail(recipient, notificationContext)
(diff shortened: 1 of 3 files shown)


swabbie-echo/src/main/kotlin/com/netflix/spinnaker/swabbie/echo/EchoNotifier.kt
```diff
@@ -46,15 +41,12 @@ class EchoNotifier(
...
-              retrySupport.retry({
-                sendEmail(recipient, notificationContext)
-              }, maxAttempts, timeoutMillis, false)
-
-              return NotificationResult(recipient, NotificationType.EMAIL, success = true)
+            return try {
+              sendEmail(recipient, notificationContext)
+              NotificationResult(recipient, NotificationType.EMAIL, success = true)
             } catch (e: Exception) {
               log.error("Failed to send notification", e)
```
aravindmd, line 48: should we log some context info for debugging if needed

jeyrschabu:
Notification tasks are placed on the queue on a tighter interval, so the chance of another node picking up ongoing work is more than likely.  This forces a node to acquire a lock on notification work before really proceeding.

In the future, I'd like to remove some indirections w/r to coordination but this would do for now.

PR #395, opened 2020-04-07
chore(java11): Compile with Java 11 (but targeting Java 8)

plumpy:
Some additional changes:

* Added a new format to `Dates.kt` because the Java 11 JRE returns nanoseconds for `LocalDateTime.now()`, so a few tests were failing. Another option is to use `LocalDateTime.now().truncateTo(MILLIS)` but this seemed more future-proof to me. Let me know if you'd prefer a different fix, however.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,9 +13,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      # Install Java 8 for cross-compilation support. Setting it up before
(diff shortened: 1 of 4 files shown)


jeyrschabu:
Lgtm, will give this a spin then merge it.

plumpy:
Thanks! FYI, aside from the change to `Dates.kt`, the rest of this has all been applied across almost all the other repos in the last two weeks (only swabbie, orca, and fiat are left)

PR #419, opened 2020-04-27
[Do Not Merge] fix(aws): Do not cache accounts which are not part of the workConfigurations

PerGon:
When doing the caching, do not cache accounts that are not part of the work configuration - it could be the case, that such accounts are not allowing the swabbie IAM role to access them - which kill swabbie with "not permitted" exceptions

### DIFF
--- a/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/config/AwsConfiguration.kt
+++ b/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/config/AwsConfiguration.kt
@@ -58,7 +58,7 @@ open class AwsConfiguration {
     accountProvider: AccountProvider,
     aws: AWS
   ): CachedViewProvider<AmazonImagesUsedByInstancesCache> {
-    return ImagesUsedByInstancesProvider(clock, accountProvider, aws)
+    return ImagesUsedByInstancesProvider(clock, workConfigurations, accountProvider, aws)
   }
 
   @Bean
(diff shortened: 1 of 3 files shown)


PerGon:
Hmmm.... wait a minute!

After digging a bit more into the code, I get the impression that, even tho Swabbie will only act on the accounts referenced in the `WorkConfiguration`, Swabbie still needs to have the whole "cloud" model of all the accounts managed by Spinnaker (in clouddriver).

E.g.: To know if a certain AMI can be cleaned up, Swabbie needs to check if *ANY* instance/server group exists that uses such AMI (not only the accounts part of the `WorkConfiguration`)

Is this example correct? And is it one of the reasons why Swabbie has to cache all cloud accounts configured in Clouddriver and not only the ones in the `WorkConfiguration`?

aravindmd:
> Hmmm.... wait a minute!
> 
> After digging a bit more into the code, I get the impression that, even tho Swabbie will only act on the accounts referenced in the `WorkConfiguration`, Swabbie still needs to have the whole "cloud" model of all the accounts managed by Spinnaker (in clouddriver).
> 
> E.g.: To know if a certain AMI can be cleaned up, Swabbie needs to check if _ANY_ instance/server group exists that uses such AMI (not only the accounts part of the `WorkConfiguration`)
> 
> Is this example correct? And is it one of the reasons why Swabbie has to cache all cloud accounts configured in Clouddriver and not only the ones in the `WorkConfiguration`?

Swabbie uses AWS api's for the read operations and uses `Clouddriver` for the deletes. Swabbie needs to have permissions to read from all accounts managed by Spinnaker, so that it can check for resources like `amis` that are shared with other accounts and not just the ones referenced in the `WorkConfiguration` and hence the need for this. Let me know if this answers your question.

PerGon:
@aravindmd thanks - that answers the question. Even tho I think it somewhat breaks the spinnaker model, where Clouddriver is (or should be) the only point of contact to cloud accounts. I was expecting Swabbie to get all the needed information from Clouddriver 🤔 
Do you think it would make sense for Swabbie to just talk to Clouddriver instead of integrating directly with cloud accounts?

aravindmd:
> @aravindmd thanks - that answers the question. Even tho I think it somewhat breaks the spinnaker model, where Clouddriver is (or should be) the only point of contact to cloud accounts. I was expecting Swabbie to get all the needed information from Clouddriver 🤔
> Do you think it would make sense for Swabbie to just talk to Clouddriver instead of integrating directly with cloud accounts?

Sorry for the delay , this completely slipped from my notifications , i think the original reason for this setup was to not overwhelm the system , there was some scalability issues at the time.

PR #434, opened 2020-05-07
chore(build): gradle 6.4

cfieber:


cfieber:
I dunno why but the PR build action wasn't kicking off, reopened as #435

PR #503, opened 2020-08-26
fix(*) Fixing cache scheduling failure

jeyrschabu:
- Moved scheduling logic to CacheStatus.kt
- cleaned out the shutdown logic. not needed
- Upped interval for cache status check
- removed the discovery logic in the task monitor agent that was killing the scheduling thread


jeyrschabu:
Merging this for testing. Will revert if needed.

PR #532, opened 2020-10-08
feat(aws): Checking for launch template references in imageHandler

jeyrschabu:
- added an in-memory cache for use in the imageHandler
- updated imageHandler to check for images used by launch templates

### DIFF
--- a/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/edda/Edda.kt
+++ b/swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/edda/Edda.kt
@@ -29,6 +29,7 @@ import com.netflix.spinnaker.swabbie.aws.images.AmazonImage
 import com.netflix.spinnaker.swabbie.aws.instances.AmazonInstance
 import com.netflix.spinnaker.swabbie.aws.launchconfigurations.AmazonLaunchConfiguration
 import com.netflix.spinnaker.swabbie.aws.launchtemplates.AmazonLaunchTemplate
+import com.netflix.spinnaker.swabbie.aws.launchtemplates.AmazonLaunchTemplateVersion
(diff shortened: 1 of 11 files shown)


jeyrschabu:
We check for references by looking at launch configs and instances in all accounts. This PR adds an additional check to make sure we account for images used by launch templates.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/images/AmazonImageHandler.kt
```diff
@@ -194,6 +196,7 @@ class AmazonImageHandler(
         setUsedByLaunchConfigurations(images, params)
         setUsedByInstances(images, params)
         setSeenWithinUnusedThreshold(images)
+        setUsedByLaunchTemplates(images, params)
```
jeyrschabu, line 199: Checks if images are referenced by launch templates. If so, the ami would not qualify for deletion.

swabbie-aws/src/main/kotlin/com/netflix/spinnaker/swabbie/aws/edda/Edda.kt
```diff
@@ -175,6 +176,23 @@ class Edda(
...
+  override fun getLaunchTemplateVersions(params: Parameters): List<AmazonLaunchTemplateVersion> {
+    val result = call(params, SERVER_GROUP) {
+      eddaService: EddaService ->
+      eddaService.getLaunchTemplateVersions()
+    } ?: emptyList()
+
+    return result.flatMap { it.versions }
+  }
+
+  data class EddaLaunchTemplaVersion(
```
jeyrschabu, line 191: This is to workaround the current edda payload. If they remove this endpoint in favor of the more traditional endpoints, this will begin to fail. left a TODO in there for when that happens :)

cfieber (approved):

PR #542, opened 2020-11-10
fix(core): Evenly distribute delete resources across the configured delete spread time divided by the total number of resource partitions

jonsie:
Previously, we had jitter across a fixed range that was added to these deletes.  This still ended up with deletes being clumped into a fairly tight window, which caused AWS rate limiting in downstream services.  This proposes to delete across an evenly distributed time window.

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
@@ -34,6 +34,7 @@ import com.netflix.spinnaker.swabbie.model.MarkedResource
 import com.netflix.spinnaker.swabbie.model.OnDemandMarkData
 import com.netflix.spinnaker.swabbie.model.Resource
 import com.netflix.spinnaker.swabbie.model.ResourceEvaluation
+import com.netflix.spinnaker.swabbie.model.ResourcePartition
 import com.netflix.spinnaker.swabbie.model.ResourceState
 import com.netflix.spinnaker.swabbie.model.Status
 import com.netflix.spinnaker.swabbie.model.Summary
@@ -83,7 +84,7 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
    * Deleted resources are produced via [DeleteResourceEvent]
    * for additional processing
    */
-  abstract fun deleteResources(markedResources: List<MarkedResource>, workConfiguration: WorkConfiguration)
+  abstract fun deleteResources(resourcePartition: ResourcePartition, workConfiguration: WorkConfiguration)
 
   /**
    * finds & tracks cleanup candidates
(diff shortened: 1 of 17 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -527,21 +528,33 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
...
+        partitions.add(ResourcePartition(markedResources = groupedMarkedResources))
       }
     }
 
-    return partitions.sortedByDescending { it.size }
+    if (partitions.isNotEmpty()) {
+      val offsetIntervalMs = configuration.getDeleteSpreadMs(dynamicConfigService) / partitions.size
+      var offsetMs = 0L
+      partitions.forEach { partition ->
+        if (offsetMs != 0L) offsetMs += offsetIntervalMs
```
dreynaud, line 552: this conditional seems odd, we want to add the `offsetIntervalMs` on every iteration, right?

jonsie: Oh wait I thought I deleted that. One moment.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -527,21 +528,33 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
...
+        partition.offsetMs = offsetMs
+      }
+    }
```
dreynaud, line 555: How about this instead?
```suggestion
      partitions.forEach { partition ->
        partition.offsetMs = offsetMs
        offsetMs += offsetIntervalMs
      }
    }
```

jonsie: Yes yes.

swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -527,21 +528,33 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
...
+      Lists.partition(it.value, configuration.itemsProcessedBatchSize).forEach { groupedMarkedResources ->
+        partitions.add(ResourcePartition(markedResources = groupedMarkedResources))
       }
     }
 
-    return partitions.sortedByDescending { it.size }
+    if (partitions.isNotEmpty()) {
+      val offsetIntervalMs = configuration.getDeleteSpreadMs(dynamicConfigService) / partitions.size
+      var offsetMs = 0L
+      partitions.forEach { partition ->
```
dreynaud, line 551: for some extra fancy pantsiness:

```suggestion
      partitions.forEachIndexed { index, partition ->
        partition.offsetMs = offsetMs + index * offsetIntervalMs
      }
```

aravindmd:
Overall approach sounds good to me , might need a couple of deletion tests to be updated.

aravindmd (approved):

PR #543, opened 2020-11-10
fix(core): Limit number of in-progress tasks submitted when deleting resources

jonsie:
Different proposal than https://github.com/spinnaker/swabbie/pull/542

This one is basically to limit the number of in-progress Orca task on a per resource type basis.

⚠️ No tests yet, build fails. ⚠️

### DIFF
--- a/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
+++ b/swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
@@ -26,6 +26,7 @@ import com.netflix.spinnaker.swabbie.events.Action
 import com.netflix.spinnaker.swabbie.events.DeleteResourceEvent
 import com.netflix.spinnaker.swabbie.events.MarkResourceEvent
 import com.netflix.spinnaker.swabbie.events.OptOutResourceEvent
+import com.netflix.spinnaker.swabbie.events.OrcaTasksStatusRefreshed
 import com.netflix.spinnaker.swabbie.events.UnMarkResourceEvent
 import com.netflix.spinnaker.swabbie.exclusions.ResourceExclusionPolicy
 import com.netflix.spinnaker.swabbie.exclusions.shouldExclude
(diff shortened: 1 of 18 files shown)


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/AbstractResourceTypeHandler.kt
```diff
@@ -490,6 +504,29 @@ abstract class AbstractResourceTypeHandler<T : Resource>(
...
 
+  /**
+   * Downstream services may be overloaded or rate limited from other services due to too many task submissions from Swabbie.
+   *
+   * If submitted in-progress tasks are above the configured threshold, wait until the task repository cache is refreshed.
+   */
+  private fun waitForTaskProcessing(taskIds: MutableList<String>, workConfiguration: WorkConfiguration) {
+    if (taskTrackingRepository.getInProgress().count { taskIds.contains(it) } >= workConfiguration.maxTasksPerBatch) {
+      latch = CountDownLatch(1)
+      latch.await()
```
jonsie, line 515: Need to test this, not totally sure if this will work -- may just need to do a `sleep` here.  Either way, a bit gross.

aravindmd: Wondering if this can be made a little simpler by setting `maxItemsToProcess` to be a difference between the `taskTrackingRepository.getInProgress().count` and `maxTasksPerBatch` , we may not have to sleep or wait then ? thoughts ?

jonsie: Problem with that is we would then only process the `maxTasksPerBatch` per cycle (I should really rename that to `maxTasksInProgressPerBatch`).  My goal is to still process the `maxItemsProcessedPerCycle` but just limit the total number of in-progress tasks (per resource type) on the Orca side.  I'm open to a simpler way though.

jonsie:
We've decided to go with #542

PR #548, opened 2020-11-13
fix(image): Fix image cleanup logic for images using launchTemplate

aravindmd:
Fix image cleanup logic for images using launchTemplate


caseyhebebrand (approved):
LGTM

PR #549, opened 2020-11-16
fix(launchTemplates): Pull aws launch template version from edda

jonsie:
- Additionally, since the total number of launch configs will diminish over time we also need to compare the total number of launch templates + launch configs when checking if images are used by launch configurations
- Removes an unnecessary sleep (which would be interrupted, throw an error, add noise to the logs) - now just check if caches are loaded as part of the normal monitor `WorkQueueManager#monitor` cycle


swabbie-core/src/main/kotlin/com/netflix/spinnaker/swabbie/work/WorkQueueManager.kt
```diff
@@ -120,18 +116,6 @@ class WorkQueueManager(
     }
   }
 
-  private fun ensureLoadedCaches() {
-    try {
-      while (!cacheStatus.cachesLoaded()) {
-        log.debug("Can't work until the caches load...")
-        Thread.sleep(Duration.ofSeconds(5).toMillis())
```
robzienert, line 127: Yikes... `Thread.sleep`...

robzienert (approved):

PR #583, opened 2022-11-04
chore(swabbie): Fix package resolution and update to Kotlin 1.6.21

jervi:
Swabbie was unbuildable due to Bintray dependencies. This PR makes the project buildable again.

The build and release scripts in `.github/workflows/` may still need som love by someone that knows the current release process...

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 korkVersion=7.150.0
-kotlinVersion=1.4.0
+kotlinVersion=1.6.21
 org.gradle.parallel=true
 spinnakerGradleVersion=8.23.0
 targetJava11=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,8 +1,8 @@
-korkVersion=7.106.0
+korkVersion=7.150.0
+kotlinVersion=1.4.0
 org.gradle.parallel=true
-spinnakerGradleVersion=8.10.1
+spinnakerGradleVersion=8.23.0
 targetJava11=true
-
 # To enable a composite reference to a project, set the
 #  project property `'<projectName>Composite=true'`.
 #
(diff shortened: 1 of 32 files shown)


jervi:
@spinnaker/oss-reviewers PTAL

gradle.properties
```diff
@@ -1,8 +1,8 @@
-korkVersion=7.106.0
+korkVersion=7.150.0
+kotlinVersion=1.6.21
 org.gradle.parallel=true
-spinnakerGradleVersion=8.10.1
+spinnakerGradleVersion=8.23.0
```
dbyron-sf, line 4: Would you mind trying the recently released 8.25.0?  I suppose we can bump it in a subsequent PR too...

gradle.properties
```diff
@@ -1,8 +1,8 @@
-korkVersion=7.106.0
+korkVersion=7.150.0
+kotlinVersion=1.6.21
```
dbyron-sf, line 2: Are we going to run into snags because other repos (e.g. [kork](https://github.com/spinnaker/kork/blob/1a53ec62e94ba42f0e2d23946dc4d71751eec353/gradle.properties#L1), [clouddriver](https://github.com/spinnaker/clouddriver/blob/6c30c181b9f92e0f3c80c5dec547b1c20315246b/gradle.properties#L6)) use kotlin 1.4.x?

gradle.properties
dbyron-sf: Actually, never mind.  There's https://github.com/spinnaker/swabbie/pull/580 which we can rebase once this goes in.

gradle.properties
jervi: We're running this version in our production accounts, I haven't seen any snags.

jervi: I'm still compiling for Kotlin 1.4

dbyron-sf: OK.  We don't use swabbie, and working builds are better than broken builds...

dbyron-sf (approved):

PR #584, opened 2023-01-03
chore(mergify): stop using strict mode in mergify

dbyron-sf:
since it's deprecated.  See https://blog.mergify.com/strict-mode-deprecation/


dbyron-sf:
@mergifyio update

PR #586, opened 2023-01-25
feat(auth): Support for spinnaker user

jervi:
The `X-SPINNAKER-USER` header will be set to the value of `ok-http-client.spinnaker-user` (defaulting to `swabbie@spinnaker.io`), and `X-SPINNAKER-ACCOUNTS` will be contain all the accounts that are configured in Swabbie. This is already supported for the Edda client, but not for calls to Echo, Front50, Orca or Clouddriver, so this PR just change the different service classes to use the same retrofit/okHttpClient as Edda and adds the `X-SPINNAKER-ACCOUNTS` header as well. These headers are needed if you are trying to clean up any protected AWS accounts.


jervi:
Before and after the change (screenshot from the Swabbie tasks overview):
image

jervi:
@spinnaker/oss-reviewers PTAL

dbyron-sf:
I'm not too familiar with swabbie, but the change looks OK.  Would love to see a test for the new functionality though.

jervi:
I didn't add tests because none of the packages I touched have a single test from before 😬

dbyron-sf (approved):

PR #666, opened 2024-05-29
fix(aws): Change default region to us-east-1

jervi:
Regional STS endpoints can be turned on or off, so there is no guarantee that `us-west-2` is enabled. However, `us-east-1` is special as it is the only regional endpoint that can't be turned off. This fixes the following error:

```
com.amazonaws.services.securitytoken.model.RegionDisabledException: STS is not activated in this region for account:01234567890. Your account administrator can activate STS in this region using the IAM Console.
```


jasonmcintosh (approved):
east-1 is ALSO the default IAM region for most STS operations.

PR #668, opened 2024-05-29
feat(swabbie): Dockerize Swabbie

jervi:


Dockerfile.compile
```diff
@@ -0,0 +1,8 @@
+FROM ubuntu:jammy
+RUN apt-get update && apt-get install -y \
+    openjdk-11-jdk \
```
jasonmcintosh, line 3: Wondering if shouldn't run on 17 vs. 11 kinda thing for builds now.  The target on the gradle plugin should still compile to 11... but can use the 17 JDK for this.

jasonmcintosh: NOTE:  This is how other projects work, so separate PR for that.

.github/workflows/release.yml
```diff
@@ -30,16 +58,106 @@ jobs:
...
+        with:
+          credentials_json: '${{ secrets.GAR_JSON_KEY }}'
+      - name: Upload halconfig profiles to GCS
+        # https://console.cloud.google.com/storage/browser/halconfig
+        # Only run this on repositories in the 'spinnaker' org, not on forks.
+        if: startsWith(github.repository, 'spinnaker/')
+        uses: 'google-github-actions/upload-cloud-storage@v2'
+        with:
+          path: 'halconfig/'
+          destination: 'halconfig/${{ steps.build_variables.outputs.REPO }}/${{ steps.release_info.outputs.RELEASE_VERSION }}'
```
jasonmcintosh, line 97: I've not looked - does swabbie HAVE halconfig stuff at all??

.github/workflows/release.yml
```diff
@@ -8,16 +8,44 @@ on:
...
+          # one branch.  And echo -n makes it appears that there's one less branch than there
+          # actually is.
+          if [ -z "$BRANCHES" ]; then
+            echo "exactly one branch required to release orca, but there are none"
+            exit 1
+          fi
+          NUM_BRANCHES=$(($(echo "$BRANCHES" | wc -l)))
+          echo "NUM_BRANCHES is '${NUM_BRANCHES}'"
+          if [ $NUM_BRANCHES -ne 1 ]; then
+            echo "exactly one branch required to release orca, but there are $NUM_BRANCHES ($BRANCHES)"
```
jasonmcintosh, line 35: This isn't orca ;) MINOR thing, just a clean it up (see it's copied from orca's release process).

jervi:
Thanks @jasonmcintosh! Not sure if any credentials or other env variables needs to be added to the repo, but hopefully everything is inherited from the org.

PR #669, opened 2024-06-27
chore(test): Adds swabbie integration tests for docker builds

jasonmcintosh:
Duplicates most of the other project integration tests with a FEW changes
* Uses valkey for redis (license concern) and uses version 7.


dbyron-sf:
Doesn't this need a chance to some github actions to actually run the tests?

jervi (approved):
Nice

jasonmcintosh:
> Doesn't this need a chance to some github actions to actually run the tests?

Yeah :) I'd not gotten all the pieces configured.  I should have probably put this as "draft" but hadn't worried too much.  I THINK it's actually in a pretty good state now for merge.

.github/workflows/build.yml
```diff
@@ -112,3 +112,8 @@ jobs:
           tags: |
             "${{ env.CONTAINER_REGISTRY }}/${{ steps.build_variables.outputs.REPO }}:${{ github.ref_name }}-latest-java11-unvalidated-ubuntu"
             "${{ env.CONTAINER_REGISTRY }}/${{ steps.build_variables.outputs.REPO }}:${{ steps.build_variables.outputs.VERSION }}-java11-unvalidated-ubuntu"
+
+      - name: Test local slim container image
```
dbyron-sf, line 116: pr.yml also needs this.

jasonmcintosh: Thanks will add that :)

jasonmcintosh: Yeah :) That's that kork config with the default locations if I recall right.  Will dig & fix here later.

jasonmcintosh: https://github.com/spinnaker/kork/blob/2b857f57c85a8c821e35da68c1207fbf00cb335d/kork-config/src/main/java/com/netflix/spinnaker/kork/boot/DefaultPropertiesBuilder.java#L32C1-L33C1 this was what was thinking.  Checking on the config here.

jasonmcintosh: FYI fixed this with latest PR, but working through all the "clouddriver isn't available" as part of the integration tests.  Got some more work to do on the downstream deps.

PR #687, opened 2024-11-19
chore(upgrades): Update OS to latest supported releases

jasonmcintosh:


swabbie-aws/src/test/java/com/netflix/spinnaker/swabbie/aws/exclusions/AmazonTagExclusionPolicyTest.kt
```diff
@@ -140,26 +139,7 @@ object AmazonTagExclusionPolicyTest {
...
-        )
       )
-    )
-
-    clock.incrementBy(Duration.ofDays(10))
-
-    resources.filter {
-      subject.apply(it, emptyList()) == null
-    }.let { filteredResources ->
-      filteredResources.size shouldMatch equalTo(1)
```
dbyron-sf, line 160: I see this [failed](https://github.com/spinnaker/swabbie/actions/runs/11921265714/job/33225258742) with:
```
AmazonTagExclusionPolicyTest > should handle high precision date times() FAILED
    java.lang.AssertionError: expected: a value that is equal to 1
    but was: 2
        at com.natpryce.hamkrest.assertion.Assertion.that(assert.kt:17)
        at com.natpryce.hamkrest.assertion.Assertion.that$default(assert.kt:14)
        at com.natpryce.hamkrest.assertion.Assert.assertThat(assert.kt:45)
        at com.natpryce.hamkrest.should.Should.shouldMatch(should.kt:15)
        at com.netflix.spinnaker.swabbie.aws.exclusions.AmazonTagExclusionPolicyTest.should handle high precision date times(AmazonTagExclusionPolicyTest.kt:160)
```
but it doesn't seem correct to totally remove the assertions.  Do you know why it failed?

jasonmcintosh: Yeah that's because i copied & pasted the above test & realized today that test seemed to depend upon the date comparison.  AKA the test was just a copy/paste to test serialization of the date string but I included all the OTHER logic without really looking at it so i stripped it all out JUST for the serialization aspects.

dbyron-sf: But this test needs at least one assertion, right?

jasonmcintosh: Technically no ;) All we're doing is testing the serialization - assertions are NOT technically required particularly if an exception is expected on an operation.  E.g. this is... almost just testing the constructor behavior passes without fialures.  I can add an assertion if we really want.

dbyron-sf: Fair enough.

dbyron-sf (approved):

